PR #13, opened 2016-02-04
Updated README.md

mikeres0:
Within the examples for using ```firebase.login({})``` the syntax used is incorrect, for example ```type: firebase.loginType.ANONYMOUS``` should be ```type: firebase.LoginType.ANONYMOUS```.


EddyVerbruggen:
Thx!

PR #16, opened 2016-03-01
return generated key

sflament:
change making  push function returning the generated key
remaining Issue : I see the correct generated key with console .log, but i don't succed to return the value with the resolve().

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -227,8 +227,11 @@ firebase.removeEventListener = function (listener, path) {
 firebase.push = function (path, val) {
   return new Promise(function (resolve, reject) {
     try {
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
I'll keep this open until you find a solution :)

sflament:
Hello Eddy,
I'm embarrassed because I do not have the skills to solve the problem. Really !

I adapted my source code to get the key generated through the result function.
It's not so easy than it could be with an immediate return code but I cope with this issue this way.

May be you want me to delete this pull request ?

EddyVerbruggen:
No worries! I'll close it and if somebody else needs this feature he can look at the work you've already done and build upon that.

PR #18, opened 2016-03-09
Corrected TypeScript definitions with firebase calls returning Promises

bpmerkel:


firebase.d.ts
```diff
@@ -91,11 +91,11 @@ declare module "nativescript-plugin-firebase" {
...
     export function init(options : InitOptions): Promise<any>;
     export function login(options : LoginOptions): Promise<LoginResult>;
-    export function createUser(options : CreateUserOptions): string;
-    export function push(path: string, value: any): void;
-    export function setValue(path: string, value: any): void;
-    export function remove(path: string): void;
-    export function query(onValueEvent: (data: FBData) => void, path: string, options: QueryOptions): void;
-    export function addChildEventListener(onChildEvent: (data: FBData) => void, path: string): void;
-    export function addValueEventListener(onValueEvent: (data: FBData) => void, path: string): void;
+    export function createUser(options: CreateUserOptions): Promise<any>;
```
EddyVerbruggen: That's better indeed, thanks! Would it be even better to use `Promise<string>` in this case?

EddyVerbruggen: Fixed and merged, thanks!

PR #20, opened 2016-03-12
Return generated key during push for iOS and android

Lampei:
Updated push for ios and android to return the key created.  Returned key in an object to be consistent with other resolve methods throughout the code. Incorporated other commit for android code, but added the key-in-object consistency to that also.


EddyVerbruggen:
Excellent idea, thanks for the changes!

PR #24, opened 2016-03-27
Reset password feature

stevetayloruk:
Hi, I've added the resetPassword() feature for both iOS and Android.


EddyVerbruggen:
Super! Very nice Easter present, thank you! :rabbit:

PR #27, opened 2016-04-08
Added update method to list of available commands

Lampei:
I needed to be able to do an update rather than just saving the entire value object, so I've created update methods in both the android and iOS code base, and call the update methods for each.  I have not tested on android, but works successfully for me in iOS.


EddyVerbruggen:
Looking good! I'll test it and add the typescript definition and doc as well :+1:

PR #28, opened 2016-04-14
Fixed all values being converted to true.

Lampei:
Previously, true and false values were being converted to String(val).  JS then converted everything to true.  This will fix that issue.


Lampei:
I did not touch the Android version as that may be working correctly (I have not tested that side).

EddyVerbruggen:
Wow, great find & fix, thanks!

PR #31, opened 2016-04-29
Implement changePassword feature

guillaume-roy:


EddyVerbruggen:
Very well done and I was happy to see you also update the readme and TS definitions!

I've granted you write access to the repo so if you ever need something else you can go right ahead and add it.

Thanks for contributing 👍

guillaume-roy:
Thank you ! :+1:

PR #77, opened 2016-07-18
Allow users to pass scope for Facebook Authentication

sean-perkins:
Checks for an optional `scope` parameter on the `arg` object of the login method. Sets the defaults to `public_profile` and `email`, but allows a user to request any required scope for their application.


EddyVerbruggen:
Thanks, looks great!

PR #101, opened 2016-08-10
Post Install Concept

sean-perkins:
Current Situation:
In a multi-developer environment, each developer has to manually configure their `Podfile` and `include.gradle` file to include any of the add-on micro-services that firebase offers. This isn't ideal, because often times users are not aware of these steps when being added to an existing project.

Proposed Solution:
This pull-request contains a post-install script that will run after every installation of the firebase plugin. It prompts the user if they are using iOS and/or Android and then walks the user through the available micro-services. Once the prompts complete, the script will overwrite the `Podfile` and `include.gradle` file with the base dependencies, as well as the selected micro-services. Items that they prompted "n" (no) for, will be commented out as they were previously. 

I'm open feedback and/or changes. I just ran into this issue with my own plugins and found that this was an effective solution for my team environment. 

Concept:
#

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,0 +1,197 @@
+var fs = require('fs');
+var prompt = require('prompt');
+
+// Default settings for using ios and android with Firebase
+var usingiOS = false, usingAndroid = false;
+
+// The directories where the Podfile and include.gradle are stored
+var directories = {
+    ios: './platforms/ios',
+    android: './platforms/android'
+}
+
+console.log('NativeScript Firebase Plugin Installation');
+prompt.start();
+askiOSPrompt();
+
+/**
+ * Prompt the user if they are integrating Firebase with iOS
+ */
+function askiOSPrompt() {
+    prompt.get({
+        name: 'using_ios',
+        description: 'Are you using iOS (y/n)',
+        default: 'y',
+    }, function (err, result) {
+        if (err) {
+            return console.log(err);
+        }
+        if (result.using_ios && result.using_ios.toLowerCase() === 'y') {
+            usingiOS = true;
+        }
+        askAndroidPrompt();
+    });
+}
+
+/**
+ * Prompt the user if they are integrating Firebase with Android
+ */
+function askAndroidPrompt() {
+    prompt.get({
+        name: 'using_android',
+        description: 'Are you using Android (y/n)',
+        default: 'y',
+    }, function (err, result) {
+        if (err) {
+            return console.log(err);
+        }
+        if (result.using_android && result.using_android.toLowerCase() === 'y') {
+            usingAndroid = true;
+        }
+        if(usingiOS || usingAndroid) {
+            promptQuestions();
+        }
+    });
+}
+
+/**
+ * Prompt the user through the configurable firebase add-on services
+ */
+function promptQuestions() {
+    prompt.get([{
+        name: 'remote_config',
+        description: 'Are you using Firebase RemoteConfig (y/n)',
+        default: 'n',
+    }, {
+        name: 'messaging',
+        description: 'Are you using Firebase Messaging (y/n)',
+        default: 'n'
+    }, {
+        name: 'storage',
+        description: 'Are you using Firebase Storage (y/n)',
+        default: 'n'
+    }, {
+        name: 'facebook_auth',
+        description: 'Are you using Facebook Authentication (y/n)',
+        default: 'n'
+    }, {
+        name: 'google_auth',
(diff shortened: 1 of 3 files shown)


scripts/postinstall.js
```diff
@@ -0,0 +1,197 @@
...
+function askAndroidPrompt() {
+    prompt.get({
+        name: 'using_android',
+        description: 'Are you using Android (y/n)',
+        default: 'y',
+    }, function (err, result) {
+        if (err) {
+            return console.log(err);
+        }
+        if (result.using_android && result.using_android.toLowerCase() === 'y') {
```
EddyVerbruggen: You could use the `isSelected` convenience function here as well.

EddyVerbruggen:
Man, I totally love it! ❤️

I only have a few trivial remarks but without those I think it's already good enough to merge, so your call if you want to add anything before I merge it.

- Can you add a message at the end of the script telling folks how to re-run the script if they made a mistake or want to add a microservice at a later time?
- Can you change the 'Are you using Facebook Authentication' prompt to 'Are you using Firebase Facebook Authentication'? I totally misread that line as 'Are you using Firebase Authentication' since it visually aligns so well with the previous lines :)
screen shot 2016-08-11 at 11 58 38
- Same for Google.


Btw, please don't remove the Podfile and include.gradle files for now (which you haven't, so that's cool) because I'm quite sure it will break builds for Telerik Platform users. I can live with the duplicate maintenance for now.

Also, seeing this work so well do you think it would be possible to automate adding [those 2 lines](https://github.com/EddyVerbruggen/nativescript-plugin-firebase#open-platformsandroidbuildgradle) to the `platforms/android/build.gradle` script as a post install step? That would require the Android platform to be already added, but that's probably mostly the case.

I will grant you commit access so you can work in this repo directly if you please (fi. for hotfixes to the script, if required).

ullalaaron:
This is Amazing!

sean-perkins:
@EddyVerbruggen Thanks for the great feedback! I'll be digging into your suggested improvements now 👍

sean-perkins:
@EddyVerbruggen The suggested improvements have been implemented. This branch should be ready to merge.

Thanks!!

jlooper:
I'm so excited about this 👯 👯 👯 👯 👯

scripts/postinstall.js
```diff
@@ -0,0 +1,197 @@
...
+    android {
+        productFlavors {
+            "fireb" {
+                dimension "fireb"
+            }
+        }
+    }
+
+    repositories {
+        mavenCentral()
```
White0ut: With how gradle works, gradle selects the first repository that contains the required dependency and uses that.  per [this release](http://www.technotalkative.com/android-studio-migration-maven-central-jcenter/) Android Studio migrated their default repository to jcenter.

jcenter is a superset of mavenCentral with the added bonus of serving their content over HTTPS and a CDN! 

My recommendation would be to place jcenter first so at least the repositories will default to this preference. The more extreme being that since it's a superset, removing mavenCentral shouldn't have any issues.

EddyVerbruggen:
Great, Great work!

I'll merge this tomorrow, so if anyone want to add (more) comments go right ahead.

@White0ut, also thanks for your suggestion, I'll be keeping that in mind.

EddyVerbruggen:
Hey @sean-perkins do you agree we can remove the 'application package' step since NativeScript added it (recently) by default to [in 2.1.0](https://github.com/NativeScript/android-runtime/issues/454). It gets written to `app/App_Resources/Android/app.gradle` in the exact same spot of the gradle hierarchy as the postinstall script would.

vekexasia:
@sean-perkins great job. I was in the process of doing something similar. Thank god I ended up here.

@EddyVerbruggen when do you think it will be merged?

EddyVerbruggen:
@vekexasia today

sean-perkins:
@EddyVerbruggen That sounds like a great plan to promote the later builds. I will commit the removal in a second.

EddyVerbruggen:
Merged! 🎉 

I'll tweak the docs where needed.

jlooper:
@EddyVerbruggen is this documented? @sean can we make a lovely NativeScriptSnack of your video? wonderful work!

EddyVerbruggen:
Yes it is

sean-perkins:
@jlooper Do you want a write-up of the process and concept of post-install or specifically the firebase post-install process for this plugin?

jlooper:
@sean I was just thinking that a video of the exact process of the post-install process for this plugin would be really good to include on the snacks site. In terms of docs, I was looking for it on the homepage of the readme but maybe missed it.

Issue #95, opened 2016-08-03
Error initializing firebase


PedroAnibarro1:
Hi,

The plugin throws this error randomly. 

     firebase.init error: Error: java.lang.NullPointerException: Argument 'applicationContext' cannot be null

Sometimes it works, sometimes it doesn't. Any idea?

Update:

This is what is being shown on screen

![screenshot_2016-08-03-15-57-49](https://cloud.githubusercontent.com/assets/8890152/17380102/5ab107d6-5993-11e6-9f60-607a44e61ee5.png)

I noticed that if I remove `firebase.addChildEventListener(this.onChildEvent, "/menu");` the app works (obviously without that feature, but at least it opens).

PedroAnibarro1:
This is the whole error message on phone:

com.tns.NativeScriptException: 
Calling js method onCancelled failed

Error: java.lang.AbstractMethodError: abstract method "void com.google.firebase.database.ChildEventListener.onCancelled(com.google.firebase.database.DatabaseError)"
    com.tns.Runtime.callJSMethodNative(Native Method)
    com.tns.Runtime.dispatchCallJSMethodNative(Runtime.java:861)
    com.tns.Runtime.callJSMethodImpl(Runtime.java:726)
    com.tns.Runtime.callJSMethod(Runtime.java:712)
    com.tns.Runtime.callJSMethod(Runtime.java:693)
    com.tns.Runtime.callJSMethod(Runtime.java:683)
    com.tns.gen.com.google.firebase.database.ChildEventListener.onCancelled(ChildEventListener.java:38)
    com.google.android.gms.internal.zzahh.zza(Unknown Source)
    com.google.android.gms.internal.zzajf.zzctc(Unknown Source)
    com.google.android.gms.internal.zzajk$1.run(Unknown Source)
    android.os.Handler.handleCallback(Handler.java:815)
    android.os.Handler.dispatchMessage(Handler.java:104)
    android.os.Looper.loop(Looper.java:194)
    android.app.ActivityThread.main(ActivityThread.java:5643)
    java.lang.reflect.Method.invoke(Native Method)
    java.lang.reflect.Method.invoke(Method.java:372)
    com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:960)
    com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)
"<unknown>, line: 0, column: -1

StackTrace: 

[...]

PR #123, opened 2016-09-03
(Bugfix Android) added onCancelled callback to child event listeners

colinskow:
This fixes `calling js method onCancelled failed` error as described in #95. It happens when a user logs out and is no longer authenticated, but has listeners on authenticated data. The listeners fail, but the app crashes since there is no "onCancelled" handler.

With this fix the app will no longer crash if the listener fails.


EddyVerbruggen:
Ha, smart move, thanks! Btw, is there a similar issue on iOS?

colinskow:
I haven't tested on iOS yet, but that is one of my next steps.

PR #145, opened 2016-09-30
Javascript implementation of .setValue and .push for android cannot handle primitive types

samuelagm:
The javascript .setValue implementation only accepts java.util.HashMap, this produces a weird behaviour of turning a string to an array of characters on firebase, and doesn't send any data to firebase when integers or booleans are used, probably due to an exception in the toHashMap function. 

This is the culprit:
`firebase.instance.child(path).setValue(firebase.toHashMap(val));`

So I just want to check if the object is primitive then represent it with appropriate Java type before passing it to setValue.


EddyVerbruggen:
Hi! I appreciate that little change, but your PR is __massive__. Would you be able to extract only the relevant part? 😄

samuelagm:
Ouch..

PR #148, opened 2016-10-03
 added a toValue function that handles primitive types on android

samuelagm:
Javascript implementation of firebase.push/setValue on android has issues with primitive types; strings are stored in firebase db as an array of characters, booleans don't get to firebase db at all, e.t.c. This is as a result of toHashMap being called on values passed to .push and .setValue regardless of their type.
I've added the ability to handle them.

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -72,6 +72,30 @@ firebase.toHashMap = function(obj) {
   return node;
 };
 
+firebase.toValue = function(val){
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
I must admit I didn't test this case, so thank you very much for adding this fix @samuelagm!

samuelagm:
Don't mention.

Regards.

Issue #179, opened 2016-10-17
Error TS2656: Exported external package typings file '[...]/node_modules/nativescript-plugin-firebase/firebase.d.ts' is not a module.


giordy:
Hello,

I floowed the instructions in the README file but I keep getting this error:

`app/app.component.ts(13,27): error TS2656: Exported external package typings file '/home/gio/GIT/xyz/node_modules/nativescript-plugin-firebase/firebase.d.ts' is not a module. Please contact the package author to update the package definition.`

Could it be that some intervention is needed on `firebase.d.ts`, as the error says?

Thanks

ullalaaron:
I've also encountered the same issue.
As a temporary fix I removed the `typings` line in the `package.json` of the plugin.
Eddy, do you have more insight on this?

giordy:
I just found out that if Typescript is updated to version 2.0.3 the error disappears.

Perhaps it could be a good idea to add a disclaimer in the README? The stub apps coming from NativeScript include version 1.8.10 of TypeScript (at least the ones I used, perhaps now this has changed).

EddyVerbruggen:
@giordy That's ok, are you willing to do a PR?

giordy:
@EddyVerbruggen sure, working on it

PR #184, opened 2016-10-17
Fix typings, addresses issue #179

giordy:
This PR should fix issue #179 

I followed the handbook that you can find at: https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html
More precisely, this is the template that seems to fit the project's use case: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html

I also renamed the typings file from `firebase.d.ts ` to `index.d.ts`, as suggested in the documentation.

I tested this with TypeScript 1.8 and 2.0 and in both cases the app compiled and loaded correctly and the firebase framework loaded properly (I got the "firebase.init done" message from the example in the README).

I couldn't test more elaborated use cases.


giordy:
@EddyVerbruggen ha! didn't notice it ^_^'
In the tests I didn't change it however it seems like a good idea to update it

EddyVerbruggen:
Thanks for confirming that, I'll change that as well, thx!

PR #198, opened 2016-11-01
Some automations for Android

PanayotCankov:
- Apply plugin for android handled in plugin's include.gradle
 - Initial questionary answers saved to firebase.nativescript.json so it can roundtrip source control without requiring the answers to be replayed. Will survive CI
 - Installed hook that would copy google-services.json from app/App_Resources/Android to platforms/android so it can roundtrip source control and CI
 - Added some missing pieces in the MESSAGING.md with some android services required to handle FCM while app is in background
 - Fixed setting application.onLaunch with adding application.on("launch" so multiple plugins don't defy each-other

Adding buildscript>dependencies>classpath "com.google.gms:google-services:3.0.0" still has to be done by hand in platforms/android or in hook. We ({N}) will have to modify the CLI to make a copy of build.gradle in app/App_Resources/Android/build.gradle in future.

### DIFF
--- a/docs/MESSAGING.md
+++ b/docs/MESSAGING.md
@@ -19,8 +19,28 @@ Open `app/App_Resources/iOS/Info.plist` and add this to the bottom:
 ```xml
 <key>UIBackgroundModes</key>
 <array>
-	<string>remote-notification</string>
+    <string>remote-notification</string>
 </array>
+```
+
+For Android you will not get the title and body if the notification was received while the application was in background, but you will get the *data* payload.
+Add the following services in the `app/App_Resources/Android/AndroidManifest.xml` to enable advanced FCM messaging:
+```
+<manifest ... >
+    <application ... >
+        ...
+        <service android:name="org.nativescript.plugins.firebase.MyFirebaseInstanceIDService">
+            <intent-filter>
+                <action android:name="com.google.firebase.INSTANCE_ID_EVENT"/>
+            </intent-filter>
+        </service>
+        <service android:name="org.nativescript.plugins.firebase.MyFirebaseMessagingService">
+            <intent-filter>
+                <action android:name="com.google.firebase.MESSAGING_EVENT"/>
+            </intent-filter>
+        </service>
+    </application>
+</manifest>
 ```
 
 #### Provisioning hell
--- a/platforms/android/include.gradle
+++ b/platforms/android/include.gradle
@@ -38,3 +38,5 @@ dependencies {
 //    compile "com.google.android.gms:play-services-auth:9.6.+"
 
 }
+
+apply plugin: "com.google.gms.google-services"
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -136,6 +136,7 @@ function promptQuestionsResult(result) {
     }
     if(usingAndroid) {
         writeGradleFile(result);
+        writeGoogleServiceCopyHook();
     }
     console.log('Firebase post install completed. To re-run this script, navigate to the root directory of `nativescript-plugin-firebase` in your `node_modules` folder and run: `npm run config`.');
 }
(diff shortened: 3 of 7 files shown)


scripts/postinstall.js
```diff
@@ -11,8 +12,40 @@ var directories = {
...
+    try {
+        config = JSON.parse(fs.readFileSync(pluginConfigPath));
+    } catch(e) {
+        console.log("Failed reading config at " + pluginConfigPath);
+        console.log(e);
+        config = {};
+    }
+}
+
+if (process.argv.indexOf("config") == -1 && fs.existsSync(pluginConfigPath)) {
```
EddyVerbruggen: I think this should read `process.argv.indexOf("config") > -1`, right?

docs/MESSAGING.md
```diff
@@ -19,10 +19,30 @@ Open `app/App_Resources/iOS/Info.plist` and add this to the bottom:
...
 <array>
-	<string>remote-notification</string>
+    <string>remote-notification</string>
 </array>
 ```
 
+For Android you will not get the title and body if the notification was received while the application was in background, but you will get the *data* payload.
+Add the following services in the `app/App_Resources/Android/AndroidManifest.xml` to enable advanced FCM messaging:
+```
+<manifest ... >
```
EddyVerbruggen: Thanks for this! Do you think we'd be able to add this to the postinstall script? Not as part of this PR, just would like to know if you think it would be a wise thing to do.

platforms/android/include.gradle
```diff
@@ -38,3 +38,5 @@ dependencies {
 //    compile "com.google.android.gms:play-services-auth:9.6.+"
 
 }
+
+apply plugin: "com.google.gms.google-services"
```
EddyVerbruggen: Sadly on my machine this leads to:

```
* Where:
Script '/Users/eddyverbruggen/sandboxes/_remove/firebasepanayot/platforms/android/configurations/nativescript-plugin-firebase/include.gradle' line: 42

* What went wrong:
A problem occurred evaluating script.
> Plugin with id 'com.google.gms.google-services' not found.
```

I'm running tns 2.3.0.. perhaps this changed in a more recent version?

Still applying the plugin at the bottom of `platforms/android/build.gradle` works fine though.

scripts/postinstall.js
PanayotCankov: No, there are now two npm scripts in the plugin now. One of them is the postinstall that will check for configuration json and does the configuration. The other is "config" which is run using `npm run config` in the plugin folder and provides the "config" arg to the same script the post install is running. It will force the script to ignore the configuration.json and re-ask the configuration questions.

docs/MESSAGING.md
PanayotCankov: I think it is ok for the plugins to require some modifications to be done by hand, as long as these modifications can be kept in source control and are well documented.

If the changes happen to be in platforms/* then it would be best if there is an automated way to apply them (using hooks?) so the app can roundtrip source control and continuous integration.

platforms/android/include.gradle
PanayotCankov: Could you double check if it works when you add by hand this in platform/android:
```
  dependencies {
    classpath "com.android.tools.build:gradle:X.X.X"
    classpath "com.google.gms:google-services:3.0.0"
  }
```

Without the:
```
apply plugin: "com.google.gms.google-services"
```

scripts/postinstall.js
EddyVerbruggen: Ah, yes I see what you mean 👍

platforms/android/include.gradle
EddyVerbruggen: Indeed, adding `apply plugin: "com.google.gms.google-services"` to the bottom is no longer required. Cool!

I see what you mean with the after prepare hook but I think it's worth the brittleness to include it in the plugin. Worst case scenario the developer still needs to add the classpath by hand but generally this would make installation lots easier.

docs/MESSAGING.md
EddyVerbruggen: 👍

EddyVerbruggen:
Merged! Much much better now, thanks so much!

EddyVerbruggen:
Btw, still considering that https://github.com/NativeScript/nativescript-marketplace-demo/blob/push-notifications/hooks/after-prepare/firebase.build.gradle.js script for further automation.

EddyVerbruggen:
@PanayotCankov Decided to add your "fragile" gradle script. I'll see how it goes, but it's a *major* improvement that no manual config is required anymore.

jlooper:
Look forward to trying it!

PanayotCankov:
The worst case scenario is the naive lookup for the correct place to insert the dependency fails and puts it in a totally wrong place rendering the build script syntactically incorrect making the android build fail miserably with cryptic errors hard to reason about. :)

Since you merged it I will probably remove it from the https://github.com/NativeScript/nativescript-marketplace-demo

Good news is since we have automated build should something go wrong we will find out in a timely fashion.

PR #222, opened 2016-11-22
Let iOS handle push notifications after the app was killed in the springboard

PanayotCankov:
And maybe fixes problems when the init() is called before application.start() and the iOS throws "cannot call app.whatever" as the app is null.

### DIFF
--- a/firebase.ios.js
+++ b/firebase.ios.js
@@ -17,8 +17,38 @@ firebase._addObserver = function (eventName, callback) {
 
 firebase.addAppDelegateMethods = function(appDelegate) {
 
+  function handleRemoteNotification(app, userInfo) {
+    var userInfoJSON = firebase.toJsObject(userInfo);
+    var aps = userInfo.objectForKey("aps");
+    if (aps !== null) {
+      var alert = aps.objectForKey("alert");
+      if (alert !== null && alert.objectForKey) {
+        userInfoJSON.title = alert.objectForKey("title");
+        userInfoJSON.body = alert.objectForKey("body");
+      }
+    }
+
+    firebase._pendingNotifications.push(userInfoJSON);
+    if (app.applicationState === UIApplicationState.UIApplicationStateActive) {
+      // If this is called from applicationDidFinishLaunchingWithOptions probably the app was dead (background)
+      userInfoJSON.foreground = true;
+      if (firebase._receivedNotificationCallback !== null) {
+        firebase._processPendingNotifications();
+      }
+    } else {
+      userInfoJSON.foreground = false;
+    }
+  }
+
   // we need the launchOptions for this one so it's a bit hard to use the UIApplicationDidFinishLaunchingNotification pattern we're using for other things
   appDelegate.prototype.applicationDidFinishLaunchingWithOptions = function (application, launchOptions) {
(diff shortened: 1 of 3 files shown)


EddyVerbruggen:
Tested iOS push notifications while app is open, in the background, and killed. The notification is always shown and sent to the in-app handler 😀👍👍👍👍

csaggo:
Can you share the code-snippets where firebase is configured, and how exactly the notification json to firebase looks. i updated to 3.8.4 but it still on ios  works for me in only in foreground and first app start.
when app is killed first time it just wont receive any massages again neither while killed nor when running again :(

PanayotCankov:
This is the app: https://github.com/NativeScript/nativescript-marketplace-demo
The setup:
 - [app/App_Resources/iOS/<app-name>.entitlements](https://github.com/NativeScript/nativescript-marketplace-demo/blob/master/app/App_Resources/iOS/nativescriptmarketplacedemo.entitlements)
 - [app/App_Resources/iOS/build.xcconfig](https://github.com/NativeScript/nativescript-marketplace-demo/blob/master/app/App_Resources/iOS/build.xcconfig#L10)
 - [app/App_Resources/iOS/Info.plist (app background mode for notifications)](https://github.com/NativeScript/nativescript-marketplace-demo/blob/master/app/App_Resources/iOS/Info.plist#L69)
It uses a fork of nativescript-plugin-firebase that disables the copy of the keychain sharing.

Here the app is initialized:
 - [On app start the firebase is required](https://github.com/NativeScript/nativescript-marketplace-demo/blob/master/app/app.ts#L16)
 - [In iOS init is called after the intro animations complete the first time here](https://github.com/NativeScript/nativescript-marketplace-demo/blob/master/app/views/main-page/main-page.ts#L126)
 - [If the animations were played allready in iOS and for Android init is here](https://github.com/NativeScript/nativescript-marketplace-demo/blob/master/app/common/firebase.ts#L13)

[CI sets the `const enabled = false;` to true](https://github.com/NativeScript/nativescript-marketplace-demo/blob/master/app/common/firebase.ts#L7) upon build and [sets up the api keys in hte .plist here](https://github.com/NativeScript/nativescript-marketplace-demo/blob/master/app/App_Resources/iOS/GoogleService-Info.plist)

The notifications displayed in the app consume the data from the push notification, I couldn't get the title and body in Android but found issues for ppl having problems with that in pure native apps too.

csaggo:
thank you very much! i have a workaround where i use firebase on android and push-notification plugin on ios, but obviously would prefer to have all in firebase. gonna work through your example!

BTW: i do get the data from android when sending fcm push like so:
"data" : {
                "title" : "some title",
                "message" => "some msg",
                "sound" : "default",
                "smallIcon" : "push",
                "largeIcon" : "push"
            }

in onMessageReceivedCallback data.message is correctly set. put i have to use data instead of notification and message instead of body and so on. works in back and foreground.

PanayotCankov:
Yep, that's what I am doing too. Adding inAppTitle and inAppBody in the data.
There was one more tricky part for iOS if I remember correctly, If you send the notification from the firebase console, scroll down to the "Advanced" part, when you expand the area, there is a title that must be filled.

PR #249, opened 2016-12-29
Added updateProfile method

abinici:
Hi,

The `updateProfile` method was missing, which was needed for updating the user-properties called  `displayName` and `profileURL`. I have implemented this method for both iOS and Android. And I have also verified that this method works on both platforms.

Please note, that in the latest version of Firebase Android SDK (9.8), a bug was introduced which causes displayName and photoUrl to always be null when fetched through `getCurrentUser()`:

https://github.com/firebase/FirebaseUI-Android/issues/409

/Ayhan

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -237,6 +237,14 @@ export interface CreateUserOptions {
 }
 
 /**
+ * The options object passed into the updateProfile function.
+ */
+export interface UpdateProfileOptions {
+    displayName: string;
+    photoURL: string;
+}
+
+/**
  * The options object passed into the resetPassword function.
  */
 export interface ResetPasswordOptions {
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -237,6 +237,14 @@ export interface CreateUserOptions {
 }
 
 /**
+ * The options object passed into the updateProfile function.
+ */
+export interface UpdateProfileOptions {
+    displayName: string;
```
EddyVerbruggen: I think these properties should be optional

abinici: You are right, both properties should be optional.

abinici: Yeah, that would be nice, because I'm currently using this project for a nativescript app and I'm sure I'll be contributing some more to this project during the app-development.

EddyVerbruggen: Awesome, done. Good luck with your app!

abinici: Thx :)

ParVisual:
Hi, is there a way to add displayName to the CreateUserOptions interface?

I need to save the user name when registering.

Thanks.

EddyVerbruggen:
@ParVisual [Here's the doc for the iOS method](https://firebase.google.com/docs/auth/web/manage-users#create_a_user)... doesn't look like that's possible.

But you can send it to the `updateProfile` function.

ParVisual:
@EddyVerbruggen thanks!

First create a user, then, on success: update profile of the created user.

PR #262, opened 2017-01-15
Adding ability to (un/)subscribe to topics at firebase

markosko:
Tested and works.
Have done only for android as i don't have Mac so can't test iOS but it's almost same way at iOS part.

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -1506,7 +1506,7 @@ firebase.subscribeToTopic = function(topicName){
   });
 };
 
-firebase.unsubscribeToTopic = function(topicName){
+firebase.unsubscribeFromTopic = function(topicName){
   return new Promise(function (resolve, reject) {
     try{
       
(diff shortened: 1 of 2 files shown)


rcluan:
Hi, @markosko . In your unsubscribe method shouldn't be unsubscribeFromTopic() instead of unsubscribeToTopic() when using com.google.firebase.messaging.FirebaseMessaging.getInstance()?

markosko:
Quite possible as i copied some things between lines :D i will change it :)

rcluan:
I've tested and it's working like a charm both subscribe and unsubscribe methods. =D well done

EddyVerbruggen:
Cool! Looks good to me as well. Let's merge this, or do you want to add a bit of documentation first? I can do that if you don't feel like it ;)

markosko:
Dont mind to do some docs, but can someone later pull up ios side so that we have both platforms in sync ? :) tommorow I can add docs

EddyVerbruggen:
Sweet, I'll take care of iOS as I'm planning on doing some maintenance work for this plugin in the next few days anyway.

markosko:
and add it to messaging docs part right ?

markosko:
and one more thing @EddyVerbruggen should I add info to docs about only android will be temporary available or no need for that ?

EddyVerbruggen:
@markosko I'll look at it now and take care of any docs changes. Thanks!!

Issue #146, opened 2016-09-30
Sending Data Messages without notification key.


Edujugon:
When I send a push notification without notification key the app crashes. If I add the notification key, it works like a charm.

below is the error log I get:
`
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.google.firebase.messaging.RemoteMessage$Notification.getTitle()' on a null object reference
org.nativescript.plugins.firebase.MyFirebaseMessagingService.onMessageReceived (MyFirebaseMessagingService.java:21)
com.google.firebase.messaging.FirebaseMessagingService.zzo ()
com.google.firebase.messaging.FirebaseMessagingService.zzn ()
com.google.firebase.messaging.FirebaseMessagingService.zzm ()
com.google.firebase.iid.zzb$2.run ()
java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1113)
java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:588)
java.lang.Thread.run (Thread.java:818)
`

(In data keys there are title and body)

Should I always send Notification Messages instead of Data Message?

Thanks in advance.

EddyVerbruggen:
Thanks for that clear bugreport and the stacktrace!

I've just updated the master branch of this repo with a changed Android implementation where we can deal with `null` notifications - so after an update your workaround would no longer be needed.

Will be part of the next npm release of course.

Can you test it to make sure?

Edujugon:
Now, it seems the app doesn't crash but I don't get any notification..
Are you checking for title and body from the data payload? or  doesn't it have any of those values?

I'm sending those values (title and body) as Data Message keys...
What about if I don't send those keys at all? do you set something by default?

EddyVerbruggen:
Hmm, it's [this piece of code](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/platforms/android/libraryproject/firebase/src/main/java/org/nativescript/plugins/firebase/MyFirebaseMessagingService.java#L19-L34). What you describe should work.

Edujugon:
What I understand from your code is that try to take the title and the body from the notification.
But if there isn't a notification, it doesn't try to take the title and body from the data..

Maybe after confirm that notification is null you could check if data has those values and if so, then take them like

json.put("title", data.getTitle())
          .put("body", data.getBody());

What do you think?

[...]

PR #281, opened 2017-02-09
Fix app crash when notification structure is missing

dbenninger:
This fixes https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/146 for iOS as well


EddyVerbruggen:
Thank you, looks good!

PR #293, opened 2017-02-21
Fix postinstall when nativescript is not in the PATH

rosen-vladimirov:
In case `nativescript` is not in the PATH variable, trying to execute `nativescript --version` will fail and the rest of the postinstall will not be executed.
For example on many test environments, where different CLI versions are tested, CLI is installed in a directory and executed directly from there.
Also this breaks adding the plugin with AppBuilder CLI (and other AppBuilder clients).

So try-catch the execution.


rosen-vladimirov:
ping @EddyVerbruggen

EddyVerbruggen:
Hi @rosen-vladimirov, that's a great idea and fix. Thank you!

Issue #226, opened 2016-11-24
Cannot find how to link anonymous account to other login type


kevinklinegargar:
Hi Everyone,

There's no function to link an  email and password credential to the firebase anonymous account?

Similar of this -- http://stackoverflow.com/questions/39088876/firebase-convert-anonymous-user-account-to-permanent-account-error

kevinklinegargar:
Anyone can confirm if there's a function integrated with this plugin to link other accounts type to anonymous account?

kevinklinegargar:
@EddyVerbruggen Can you help me?

EddyVerbruggen:
The plugin should automatically link an existing account when logged on with it and authenticating with Google or Facebook.

I didn't do this for email (yet) though.

kevinklinegargar:
Thanks @EddyVerbruggen . I'll give an update here once I tested on email account type

RaphaelJenni:
The login with Google/Facebook works. But the data isn't fetched correctly. For example the name is missing. Beside that if you try to reauthenticate with this "converted" user, there is an error saying "Not currently logged in with Google". You first have to logout and login again.

EddyVerbruggen:
@RaphaelJenni Can you point out in the Firebase iOS/Android documentation how that should work? I'm not sure it's incorrect what you're seeing.

RaphaelJenni:
@EddyVerbruggen I checked the documentation and your code. It seems correct, but there is a problem with the authentication.

If I login with an anonymous user, convert the user to a google user, close the app, reopen it again and reauthenticate, the message "Not currently logged in with Google" message appears. If you try to delete the account this error appears "Deleting a user failed. Credentials too old."

Beside that, if you try to connect your account with a user who already has an account there is an error saying it cannot connect the two accounts but afterwards you are logged in as the other user and the data from the anonymous user is lost.

RaphaelJenni:
@EddyVerbruggen Can you give an status update?

EddyVerbruggen:
No update from me, over.

RaphaelJenni:
Is it planned? Cause I'm gonna need that in the near feature. I can try to fix it by my self, but no guaranty that I solve it...

EddyVerbruggen:
I try to look at every issue across many repo's. It's not possible for me to tell you when this one is up next. Please take a look if you have time, I'd be glad to review.

RaphaelJenni:
[...]

PR #303, opened 2017-03-01
On login failed, show google account modal again

RaphaelJenni:
Refer to #226 .
If the user sign in with an anonymous account and tried to link it to a already used google account, he got stuck in a loop, where he wasn't able to choose a different account. Fixed that.


EddyVerbruggen:
This looks good, thank you!

PR #309, opened 2017-03-12
Fix to return downloadUrl in response of uploadFile for iOS

shripalsoni04:
I found that in iOS, when we upload any file to storage using `firebase.uploadFile` method, in response the`url` property was not coming. 

From code of `firebase.ios.ts`, I found that currently `downloadURL` of response metadata is considered as property, but actually it is a function as mentioned at https://firebase.google.com/docs/reference/ios/firebasestorage/api/reference/Classes/FIRStorageMetadata#/c:objc(cs)FIRStorageMetadata(im)downloadURL

Just changing `downloadURL` to function call, it resolves this issue.

Let me know in case of any concerns and thanks for this very useful plugin 👍

### DIFF
--- a/firebase.ios.js
+++ b/firebase.ios.js
@@ -1540,7 +1540,7 @@ firebase.uploadFile = function (arg) {
         } else {
           resolve({
             name: metadata.name,
-            url: metadata.downloadURL.absoluteString,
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Excellent, thank you!

PR #324, opened 2017-03-22
Use Gson to convert javaObject to jsObject

SBD580:
Hi,

We had major performance issues with the plugin when working with medium json files (>50K). It took a couple of seconds (5-8 seconds) to convert the javaObject to javascript one, and it was done twice instead of just one time.

Using Gson (any other library will due I guees) with JSON.parse it take only several (200-300) milliseconds.

This has not tested fully yet and maybe it should be configurable (maybe its having a performance issue on a small size files).

Would like to know why in the first place the conversion was done this way and not with some standard library.

Shay

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -14,6 +14,8 @@ firebase._facebookAccessToken = null;
 var fbCallbackManager = null;
 var GOOGLE_SIGNIN_INTENT_ID = 123;
 
+var gson = new com.google.gson.Gson();
+
 (function() {
   if (typeof(com.google.firebase.messaging) === "undefined") {
     return;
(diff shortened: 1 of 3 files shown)


EddyVerbruggen:
Thank you! I'll look at it (hopefully) soon.

The reason it is what it is, is because of [this](https://gist.github.com/jbristowe/c89a7bcae7fc9a035ee7#gistcomment-1609699). Never gave it a second thought to do it differently but your suggestions makes much sense.

EddyVerbruggen:
I may have tried that and had issues, so beware. I use Gson all the time when working on serverside Java so I'm sure your suggestion will be pretty good.

We could also take a look at {N}'s [http module source](https://github.com/NativeScript/NativeScript/tree/master/tns-core-modules/http).

SBD580:
I believe things here a bit different - we are not just parsing JSON string, but we need to transform the DataSnapshot Java object back to JSON string and then parsing it (for some reason, Firebase Andrioid library designers chose not to expose the raw data like they did for example for their Unity library, things could have been even faster that way)

EddyVerbruggen:
This seems to work really well. I only test with small data structures so don't notice a speed difference, but the conversion looks spot on. Perhaps even fixes an issue or two..

I've added some fallback code in case users updating the plugin don't automatically pull in Gson. But that's a temporary state of course. Just don't feel like a bunch of issues where folks report `com.google.gson.Gson` is undefined.

SBD580:
Sounds good :)

PR #338, opened 2017-03-28
Return Facebook access token to login result for use with FB SDK

abhayastudios:
Return the Facebook access token to the login result, so that the developer can use it for querying the FB Graph API.

It seems to work on a Samsung Galaxy 4 Android device and the iOS simulator.


EddyVerbruggen:
Looks good! Are you up for a little change to the readme as well to help others find this little gem?

abhayastudios:
@EddyVerbruggen what you you think about this?

EddyVerbruggen:
❤️ it! Thank you 👍

PR #352, opened 2017-04-21
Fix hanging CI builds

rosen-vladimirov:
### Fix usage of plugin with npm 2
Whenever you try using the plugin with `npm 2`, it's postinstall script will fail, as it requires `fs-extra` module. However it's not been added as dependency of the plugin, so it does not exist in node_modules and `require` fails.
When npm 3 or later is used, the `fs-extra` plugin exists in the root level of `node_modules` as it is a dependency of `fs-promise`. npm 3 and later tries to flatten the dependencies tree, so `fs-extra` is moved in the root level and the `require` in the post-install script works.

In order to fix this, add the `fs-extra` as dependency of the plugin.

> NOTE: Maybe it's worth checking why we need `fs-extra`, Node.js `fs` could do the work, but I leave it for further discussions.

### Fix postinstall in CI environment
In case you are in CI environment (non-interactive terminal), the postinstall script will fail as it prompts for action (in case there's no firebase.nativescript.json). As this is one of the basic scenarios when adding a plugin (add it and try building the project), some CI or cloud builds are hanging.
So add a check if the console is interactive and use the default settings in case it is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,10 +66,11 @@
   },
   "homepage": "https://github.com/eddyverbruggen/nativescript-plugin-firebase",
   "dependencies": {
-    "xcode": "0.9.2",
-    "prompt-lite": "^0.1.1",
+    "fs-extra": "2.1.2",
     "fs-promise": "^2.0.0",
-    "nativescript-hook": "^0.2.1"
+    "nativescript-hook": "^0.2.1",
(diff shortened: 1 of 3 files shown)


EddyVerbruggen:
Thanks!

I'll push an update to npm in a minute.

PR #362, opened 2017-05-03
Add Feature Firebase Invites

yezarela:
Hi Eddy, 
I've added "Firebase Invites" feature in this PR based on your [Firebase-Invites Cordova](https://github.com/Telerik-Verified-Plugins/Firebase-Invites) plugin. But it's only support Android, since I dont have any experience in iOS :D .

Would love to have iOS implemented too :+1:

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -13,6 +13,7 @@ firebase._facebookAccessToken = null;
 
 var fbCallbackManager = null;
 var GOOGLE_SIGNIN_INTENT_ID = 123;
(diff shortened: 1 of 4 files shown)


EddyVerbruggen:
This works great! Thanks a lot. Would you like to have commit access to this repo so you'll be able to fix any issues you encounter?

I'll add docs for this and a nit more TypeScript definition support for the result objects. And I'll keep iOS support in mind as well.

Issue #342, opened 2017-04-01
iOS: Simulator crash when authenticate with firebase


Visutr:
The code works on Android emulator (shown below), but not the iOS simulator, the console says segmentation fault: 11. Is this a bug with the simulator?

Android Emu works fine:
<img src="https://cloud.githubusercontent.com/assets/20510194/24580580/7ad03488-1734-11e7-9ae0-410778f35433.png" width="300px"/>

On iOS simulator:
![screen shot 2017-04-01 at 11 43 21 pm](https://cloud.githubusercontent.com/assets/20510194/24580623/3aa3f286-1735-11e7-90cc-bf75961cc308.png)

EddyVerbruggen:
Which line makes it crash, and what happens on a real device?

Visutr:
Hello Eddy,

Line 25, firebase.login(...) is the cause to this problem. I don't have a real iOS device to test on, so I can't really confirm what happens when it runs on a real device.

EddyVerbruggen:
On line 25 there's `loginType`. What's the runtime value?

Visutr:
The `loginType` value is either `LoginType.FACEBOOK`, `LoginType.GOOGLE`, it depends on user's selection. Both failed anyway.

EddyVerbruggen:
Just to be sure, `LoginType.FACEBOOK` evaluates correctly for a string?

Visutr:
As far as I can tell `LoginType.FACEBOOK`  is one of enum element I imported from `nativescript-plugin-firebase`. I didn't quite get what you mean though, please tell me if I misunderstood.

EddyVerbruggen:
I was wondering if that enum would evaluate to its string representative at runtime. So what's the output of this?

```js
console.log("runtime value of LoginType.FACEBOOK: " + LoginType.FACEBOOK);
```

Visutr:
For FACEBOOK it's `facebook` and for GOOGLE it's `google`. 

![screen shot 2017-04-03 at 11 00 45 pm](https://cloud.githubusercontent.com/assets/20510194/24618497/740daa02-18c1-11e7-9711-b4cc624f3ed0.png)

EddyVerbruggen:
Thanks, that looks fine.

[...]

PR #369, opened 2017-05-18
Validate `currentAccessToken` before accessing its `tokenString`

Jai-Krish:
This fixes the error `null is not an object (evaluating 'FBSDKAccessToken.currentAccessToken().tokenString')` specified in the issue #342


EddyVerbruggen:
That won't hurt, thanks!

Issue #399, opened 2017-06-21
v4.0.2 fails Webpack/Angular AoT


DickSmith:
```
ERROR in ../~/nativescript-plugin-firebase/firebase.ios.js
Module parse failed: /Users/dick/src/pnp-client-mobile/node_modules/nativescript-plugin-firebase/firebase.ios.js 'return' outside of function (21:2)
You may need an appropriate loader to handle this file type.
| } catch (e) {
|   console.log(">>>> Firebase is not configured correctly; please remove your 'platforms/ios' folder, then run 'tns run ios'");
|   return;
| }
|
 @ ./vendor.ts 25:0-39
```

Removing the return resolves this.

EddyVerbruggen:
Hi Dick,

Did pre-4.0 versions also fail with webpack? Just wondering if this is something I broke recently..

Thanks,
Eddy

EddyVerbruggen:
Oh nevermind.. that's indeed brand new code.

hypery2k:
having this issue, too. Any suggestions or progress on this bug?

EddyVerbruggen:
Not yet, any suggestions on your part perhaps?

hypery2k:
Expected behaviour. Top level returns are not ES6 module compliant. So that's why it's failing.

Will add a PR for this issue as soon as i verified my AoT build with iOS

hypery2k:
So mor me it's working with my changes for iOS and AoT build.

@EddyVerbruggen: Would be great if you release an patched version of the plugin

EddyVerbruggen:
Awesome, will be part of 4.0.3 which I'll publish later today.

PR #400, opened 2017-06-21
Fix for #399 v4.0.2 fails Webpack/Angular AoT

DickSmith:
Fix for #399 v4.0.2 fails Webpack/Angular AoT


EddyVerbruggen:
Thanks, I overlooked your PR I'm afraid and have now applied #414 instead. I'm going to test both approaches locally and use the one that works best.

PR #410, opened 2017-06-27
[iOS] Fix firebaseAuth reference error

vcooley:
Fixes fireBaseAuth reference error when using custom token provider function on iOS.


EddyVerbruggen:
Thanks, LGTM!

PR #414, opened 2017-06-29
Fix AoT build on iOS

hypery2k:
wrap top-level return ES6 compliant by wrapping it in a IIFE

see #399


EddyVerbruggen:
Oh dear, that change is enormous and GitHub's diff isn't helping. All you did is wrap (almost) everything in a `(function() { ... })();`?

hypery2k:
yes ;)
If you want you can do it on your own and close this PR

EddyVerbruggen:
Thanks for confirming that. I'll merge and test it, then bump the version. Many thanks! 👍

PR #424, opened 2017-07-05
refactor: wrap `gson` and `messagingEnabled` in factory funcs

sis0k0:
The above vars shouldn't be evaluated at build time, because the mksnapshot tool will fail (android runtime is not loaded, yet). That's why we need to wrap them in factory functions and instantiate them at runtime.

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -15,13 +15,28 @@ var fbCallbackManager = null;
 var GOOGLE_SIGNIN_INTENT_ID = 123;
 var REQUEST_INVITE_INTENT_ID = 48
 
-var gson = typeof(com.google.gson) === "undefined" ? null : new com.google.gson.Gson();
+function lazy(action) {
+    var _value;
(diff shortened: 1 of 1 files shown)


sis0k0:
ping @EddyVerbruggen @PanayotCankov

EddyVerbruggen:
I take your word for it! Thank you 👍

EddyVerbruggen:
I'll release a new version tomorrow.

firebase.android.js
```diff
@@ -15,13 +15,28 @@ var fbCallbackManager = null;
...
+function lazy(action) {
+    var _value;
+    var that = this;
+
+    return function() {
+        return that._value || (that._value = action());
+    }
+}
+
+var gson = lazy(function() {
```
EddyVerbruggen: Turned out I had to add `new` here (and 4 lines below) otherwise the cached values would interfere.

PR #432, opened 2017-07-14
Update README.md to accurately reflect the state of this plugin for iOS

briandilley:


EddyVerbruggen:
I don't see how that's helpful. If you have problems share a repo anyone can clone. If there's a proper reproduction path there's very likely a way to fix it. I'm using this in an app and have helped many others getting it to work in the past. It's a bit hard for me to assist everyone all the time though.

Issue #438, opened 2017-07-19
Query : range not working when value is number


GrEg00z:
Old issue : #87 

I don't understant why, but the problem comes back for me, exactly in the same context than my previous post in issue 87 :

> Same problem here, while using this :
> 
> firebase.query(({ type, key, value }) => {
>         // .......
>       }, ref, {
>          orderBy: {
>           type: firebase.QueryOrderByType.CHILD,
>           value: 'status'
>             ......................................

So if I use number value instead of string value in **ranges** array , I get the following error : 

```
TypeError: Cannot read property 'isStopped' of undefined
	at Subscriber.error (file:///data/data/org.nativescript..../files/app/tns_modules/rxjs/Subscriber.js:100:18)
	at ZoneDelegate.invoke (file:///data/data/org.nativescript..../files/app/tns_modules/nativescript-angular/zone-js/dist/zone-nativescript.js:365:26)
	at Object.onInvoke (file:///data/data/org.nativescript...../files/app/tns_modules/@angular/core/bundles/core.umd.js:4132:37)
	at ZoneDelegate.invoke (file:///data/data/org.nativescript..../files/app/tns_modules/nativescript-angular/zone-js/dist/zone-nativescript.js:364:32)
	at Zone.run (file:///data/data/org.nativescript..../files/app/tns_modules/nativescript-angular/zone-js/dist/zone-nativescript.js:125:43)
	at file:///data/data/org.nativescript..../files/app/tns_modules/nativescript-angular/zone-js/dist/zone-nativescript.js:760:57
	at ZoneDelegate.invokeTask (file:///data/data/org.nativescript..../files/app/tns_modules/nativescript-angular/zone-js/dist/zone-nativescript.js:398:31)
	at Object.onInvokeTask (file:///data/data/org.nativescript..../files/app/tns_modules/@angular/core/bundles/core.umd.js:4123:37)
	at ZoneDelegate.invokeTask (file:///data/data/org.nativescript..../files/app/tns_modules/nativescript-angular/zone-js/dist/zone-nativescript.js:397:36)
	at Zone.runTask (file:///data/data/org.nativescript..../files/app/tns_modules/nativescript-angular/zone-js/dist/zone-nativescript.js:165:47)
	at drainMicroTaskQueue (file:///data/data/org.nativescript..../files/app/tns_modules/nativescript-angular/zone-js/dist/zone-nativescript.js:593:35)

```

[...]

Issue #328, opened 2017-03-23
Query range with START_AT fetching data as expected


NordlingArt:
I left another issue (#319) about `null`not working with `EQUAL_TO`. Now I'm trying to fetch data based on `price´property where the price is over a certain amount.

I have these articles in my database:

```json
{
  "articles": {
    "-article1": {
      "title": "Article 1",
      "price": 10
    },
    "-article2": {
      "title": "Article 2",
      "price": 50
    },
    "-article3": {
      "title": "Article 3",
      "price": 100
    }
  }
}
```

This query fetches all articles, ignoring the price range:

```js
firebase.query(snapshot => {
  console.dump(snapshot); // Logs all articles (incorrect)
}, "/articles", {
  singleEvent: true,
  orderBy: { type: firebase.QueryOrderByType.CHILD, value: "price" },
  range: { type: firebase.QueryRangeType.START_AT, value: 51 } // Should fetch article 3 only
});
```

This query works with the official Firebase Node.js (Admin):

```js
admin.database().ref("/articles").orderByChild("price").startAt(51).once("value").then(snapshot => {
  console.log(snapshot.val()); // Logs article 3 only (correct)
});
```

Also, if I pass a `0` to the `START_AT` value, the snapshot function will never fire.

EddyVerbruggen:
Does this happen on iOS, Android, or both?

NordlingArt:
@EddyVerbruggen - Can only tell for iOS. Never tried on Android.

PR #444, opened 2017-08-03
Updated how options.*.value is checked to allow booleans and numbers

ivanwitzke:
This should fix the following issues: #438 and #328


EddyVerbruggen:
Hmm, that makes a lot of sense! Thank you 👍

PR #449, opened 2017-08-10
Mark init and getRemoteConfig for profiling

PanayotCankov:
These sometimes complete async or are called in the application launched event and the times spend inside are otherwise hard to track. Wrapping these in profile will show them when running the timeline profiling. init on my android phone spends about 150ms, that's fine but I'd like to be able to keep an eye on it.

I did the changes in the GitHub UI, so please test it before merge.

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -3,6 +3,7 @@ var utils = require("utils/utils");
 var frame = require("ui/frame");
 var fs = require("file-system");
 var firebase = require("./firebase-common");
+var profile = require("profiling").profile;
 
 firebase._launchNotification = null;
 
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Thanks! Is there a chance (with a lower {N} version) that `require("profiling").profile` resolves to nothing?

PanayotCankov:
Yes. This is relatively new feature.

EddyVerbruggen:
OK thanks, I will make sure it'll work with older versions as well.

EddyVerbruggen:
4.0.6 contains a backward compatible version of this PR. Now available on npm.

PR #476, opened 2017-09-02
Moved AdMob init to firebase.init, added AdListener to showBanner()

svzi:
The initialization of the AdMob plugin takes some time with current plugin-implementation. Showing the first banner could be delayed by a few seconds. Moving AdMob init from `firebase.admob.showBanner()` to `firebase.init` speeds that up a lot.

I've also added an `AdListener` to `firebase.admob.showBanner()`, similar to the one already used in `firebase.admob.showInterstitial()`. With that we can take care when a banner isn't displayed because of an network error or something else. The current implementation didn't return those kind of errors.

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -208,6 +208,9 @@ firebase.init = function (arg) {
 
       var firebaseAuth = com.google.firebase.auth.FirebaseAuth.getInstance();
 
+      // init admob
+      com.google.android.gms.ads.MobileAds.initialize(appModule.android.context); // TODO not sure its bound to packagename.. this is from the admob-demo
+
       if (arg.onAuthStateChanged) {
         firebase.authStateListener = new com.google.firebase.auth.FirebaseAuth.AuthStateListener({
           onAuthStateChanged: function (fbAuth) {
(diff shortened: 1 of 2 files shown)


firebase.android.js
```diff
@@ -208,6 +208,9 @@ firebase.init = function (arg) {
 
       var firebaseAuth = com.google.firebase.auth.FirebaseAuth.getInstance();
 
+      // init admob
+      com.google.android.gms.ads.MobileAds.initialize(appModule.android.context); // TODO not sure its bound to packagename.. this is from the admob-demo
```
EddyVerbruggen: What if a user didn't enable AdMob? Let's make this conditional.

svzi:
Hi!

> Hi, does the fact there's no iOS changes mean you didn't notice the delay there?
No, that just mean that I'm only using Android for my project so far. 

> What if a user didn't enable AdMob? Let's make this conditional.
I would love to do that conditional, but I didn't find a way on how to detect if the user didn't enable AdMob. When you tell me how to check that, I would be happy to update my PR.

EddyVerbruggen:
Would be great if you could update the PR!
I think a check like this should suffice:

```js
  if (typeof(com.google.android.gms.ads) !== "undefined") {
    // init admob here
  }
```

svzi:
I've updated my PR. I've also moved the code some rows down in the init, I just recognized that I added it into the lines of the auth init. So I moved it down.

EddyVerbruggen:
👍 👍 👍  Thanks!

svzi:
Thank you for building and maintaining this awesome plugin! Keep up the great work! 👍

PR #501, opened 2017-10-01
Added null check for updateCallback on query

derTuca:
When using the singleEvent option on a query, there are quite a few cases where we need to retrieve the data via the promise. In some of those cases, it does not make sense to also implement a callback, so a null check allows us to just pass null instead of a stub callback.


EddyVerbruggen:
That makes a lot of sense, thank you! 👍

PR #506, opened 2017-10-03
Improved Promise handling for showInterstitial()

svzi:
Let showInterstitial() fullfil promise when the ad is closed instead of when the ad is loaded.

This way we can detect when the ad is closed and perform following actions. Before we only knew the ad was loaded, but we didn't knew if the user closed the ad already.

### DIFF
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -600,7 +600,6 @@ firebase.admob.showInterstitial = function (arg) {
       var InterstitialAdListener = com.google.android.gms.ads.AdListener.extend({
         onAdLoaded: function () {
           firebase.admob.interstitialView.show();
-          resolve();
         },
         onAdFailedToLoad: function (errorCode) {
           reject(errorCode);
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Do you think this will break app behavior for current users?

What I dislike is this means it's even less aligned with iOS.

svzi:
{quote}Do you think this will break app behavior for current users?{quote}I'm not sure about it. It could break something, when someone wants to do some stuff when the interstitial is loaded. I don't know, because from my point of view it makes more sense to do stuff, when the ad has been closed... 

{quote}What I dislike is this means it's even less aligned with iOS.{quote}I can fully understand that. My app currently doesn't run on iOS and I have no experience with native iOS development. :(

EddyVerbruggen:
Roger that, thanks for the clarifications 👍 

Let's keep this open to see if others have an opinion on when the promise should resolve.

svzi:
Good idea 👍

muratcorlu:
This PR is nearly 1 year old. Do you think it's still possible to merge? There is also another related open PR: #775

PR #510, opened 2017-10-06
Add FCM data to the data key

bpeterman:
The [docs](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/docs/MESSAGING.md) state that the data from FCM messages should be under the data key.  Currently it's just in the body of the main JSON object returned.  This causes an issue for the TypeScript usage:

>   firebase.init({
    onMessageReceivedCallback: (message: Message) => {
      console.log(`Title: ${message.title}`);
      console.log(`Body: ${message.body}`);
      // if your server passed a custom property called 'foo', then do this:
      console.log(`Value of 'foo': ${message.data.foo}`);
    }
  });

The [interface](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/firebase.d.ts#L426) restricts us from using anything other than the data key .

This PR puts everything from getData underneath the "data" key in the JSON object.


EddyVerbruggen:
Looks good to me, thank you very much!

Issue #496, opened 2017-09-21
[BUG] iOS Notifications being registered to background threads


Spenstar:
I noticed that when the cloud messaging system is registered for iOS, it is put on a background thread as seen in the image below:
![screen shot 2017-09-20 at 11 23 21 pm](https://user-images.githubusercontent.com/8906394/30680692-504c9330-9e5e-11e7-8957-3f0829a71a2a.png)

I have tried calling `firebase.init()` in different parts of the application, but still get this error.

If this could be looked into, that would be great.

### UPDATE:
After playing around with this, I have found that this happens when I build the Android version of the application. When the Android build is finished, TNS needs to rebuild the iOS app again for some reason. Once that build is finished and installed on the device, push notifications no longer work on the iOS version.

One workaround (not the most ideal solution) is to uninstall the app from the iOS device and then run `tns run ios` again. This will install a fresh app and push notifications work again.

With that in mind, I'm not sure if this is an issue with the build process of NativeScript, or if this can be fixed within the Firebase plugin.

b4rtt:
@Spenstar Hi, same problem here. But uninstall app doesn't help. My app can't get any notification message :/

rhrn:
Maybe will helpful
https://github.com/evollu/react-native-fcm/commit/a2412b8ef95ea9c0e6f295501b9114f560bf7b69

EddyVerbruggen:
Thanks @Spenstar!

PR #515, opened 2017-10-12
[FIX] iOS registerForRemoteNotifications() Now Runs on Main Thread

Spenstar:
#496 Should now be resolved thanks to the suggestion given by Roman Nesterov.


EddyVerbruggen:
That's awesome!! Thank you! I'll merge this and change the `() => {` code to plain old `function() {` syntax just to be safe. Mental note: convert this baby to TS one day...

Issue #488, opened 2017-09-12
IOS app crashes after push notification confirmation


ryanzaatari:
I have been struggling for couple of days now from a similar issue, I have the latest release of this plugin `4.1.1`

here's my package.json file

> {
  "description": "Makhzoumi Foundation public application",
  "license": "SEE LICENSE IN <your-license-filename>",
  "readme": "----NO DATA AVAILABLE----",
  "repository": "<fill-your-repository-here>",
  "nativescript": {
    "id": "com.makhzoumifoundation.general",
    "tns-android": {
      "version": "2.5.0"
    },
    "tns-ios": {
      "version": "2.5.0"
    }
  },
  "dependencies": {
    "nativescript-cardview": "^1.3.2",
    "nativescript-carousel": "^2.4.2",
    "nativescript-floatingactionbutton": "^2.2.7",
    "nativescript-intl": "^3.0.0",
    "nativescript-loading-indicator": "^2.3.2",
    "nativescript-orientation": "^1.6.1",
    "nativescript-phone": "^1.3.0",
    "nativescript-plugin-firebase": "^4.0.6",
    "nativescript-pulltorefresh": "^1.1.10",
    "nativescript-sqlite": "^1.1.7",
    "nativescript-statusbar": "^1.0.0",
    "nativescript-telerik-ui-pro": "file:///Users/ryanzaatari/Documents/Development/NativeScript/nativescript-ui-pro.tgz",
    "nativescript-theme-core": "^1.0.4",
    "nativescript-toasts": "^1.0.3",
    "tns-core-modules": "^2.5.2"
  },
  "devDependencies": {
    "babel-traverse": "6.24.1",
    "babel-types": "6.24.1",
    "babylon": "6.17.0",
    "lazy": "1.0.11"
  }
}

here's the crash report from firebase console

`Exception name: EXC_BAD_ACCESS / KERN_INVALID_ADDRESS
1
NativeScript
4389574508 + 137068
2
NativeScript
4389573848 + 136408
3
NativeScript
4389625060 + 187620
4
NativeScript
4398674648 + 9237208
5
NativeScript
4398678448 + 9241008
6
Foundation
-[__NSObserver _doit:] + 36880
7
CoreFoundation
__CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 868648
8
CoreFoundation
_CFXRegistrationPost + 865992
9
CoreFoundation
[...]

PR #520, opened 2017-10-13
#488 Fix crash after push notification confirmation

rhrn:
In `addBackgroundRemoteNotificationHandler` before `FIRAuth.auth()` `FIRApp` was not configured


EddyVerbruggen:
Looks harmless to me, thank you! 👍

PR #524, opened 2017-10-16
Fixed bugs in invite and dynamic links implementations

sserdyuk:
Hi @EddyVerbruggen, I have started testing invites and dynamic links and ended up fixing them up. Here's a PR. 

Added proper iOS 9+ handling for dynamic links
Extended returned data to include match confidence (specific for iOS)
Updated documentation

### DIFF
--- a/firebase.ios.js
+++ b/firebase.ios.js
@@ -68,8 +68,8 @@ function addBackgroundRemoteNotificationHandler(appDelegate) {
     appDelegate.prototype.applicationDidReceiveRemoteNotificationFetchCompletionHandler = function (app, notification, completionHandler) {
 
       firebase._configure();
-
-      // Pass notification to auth and check if they can handle it (in case phone auth is being used), see https://firebase.google.com/docs/auth/ios/phone-auth
+      
+            // Pass notification to auth and check if they can handle it (in case phone auth is being used), see https://firebase.google.com/docs/auth/ios/phone-auth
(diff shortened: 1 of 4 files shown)


sserdyuk:
BTW, this implements dynamicLinkFromUniversalLinkURL so #505 is not needed.

firebase.ios.js
```diff
@@ -171,6 +185,37 @@ firebase.addAppDelegateMethods = function (appDelegate) {
     };
   }
 
+  if (typeof(FIRDynamicLink) !== "undefined") {
+    appDelegate.prototype.applicationContinueUserActivityRestorationHandler = function (application, userActivity, restorationHandler) {
+      var result = false;
+      
+      if (typeof(FIRDynamicLink) !== "undefined" ) {
```
EddyVerbruggen: This seems a duplicate of line 188. If you want to change it go ahead, otherwise I'll do it after merging this awesome PR 👍

sserdyuk: Let me double check if I messed up anything while rebasing.

sserdyuk: No, it's looking good. It's the same if twice, but I did it to follow the pattern that's used with other delegates. Right now `applicationContinueUserActivityRestorationHandler` is only used for one thing, but if something else will need to be added in the future, it will be more trivial and clear.

EddyVerbruggen: Smart thinking, thanks for the explanation!

PR #526, opened 2017-10-18
Implement reload of a firebase user

### LINKED ISSUE #483: Implement reload of a firebase user (needed for email verification)
[Stack Overflow discussion](https://stackoverflow.com/q/37900447)

To get the correct value of `User.emailVerified` after a user verified his email address, we have to manually refresh the user's data from the server. Currently, the only workaround is signing the user off and on again.

Since both SDKs, Android ([FirebaseUser.reload()](https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseUser.html#reload())) and iOS ([FIRUser.reloadWithCompletion](https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Classes/FIRUser#/c:objc(cs)FIRUser(im)reloadWithCompletion:)) provide specific implementations, it shouldn't be to complex to support this feature.

witi83:
Fixes #483

### DIFF
--- a/firebase.d.ts
+++ b/firebase.d.ts
@@ -786,6 +786,8 @@ export function login(options: LoginOptions): Promise<User>;
 
 export function reauthenticate(options: ReauthenticateOptions): Promise<any>;
 
+export function reload(): Promise<void>;
+
 export function getAuthToken(option: GetAuthTokenOptions): Promise<string>;
 
 export function logout(): Promise<any>;
--- a/firebase.android.js
+++ b/firebase.android.js
@@ -1281,6 +1281,32 @@ firebase.reauthenticate = function (arg) {
   });
 };
 
+firebase.reload = function () {
+  return new Promise(function (resolve, reject) {
+    try {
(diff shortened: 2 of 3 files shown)


witi83:
Since I don't have an iOS device, I'm unfortunately not able to test this change by myself. Android is working though.

firebase.d.ts
```diff
@@ -786,6 +786,8 @@ export function login(options: LoginOptions): Promise<User>;
 
 export function reauthenticate(options: ReauthenticateOptions): Promise<any>;
 
+export function reload(): Promise<void>;
```
EddyVerbruggen: I may change this to `reloadUser` to make it more clear, but I'll make up my mind after merging this awesome PR.

EddyVerbruggen:
Your code looks great, iOS as well! I'll merge and test this.

Thank you very much for your contribution! 👍 🥇

firebase.d.ts
Spenstar: I second the idea of renaming to make the function more clear 👍🏼

PR #527, opened 2017-10-18
Added constant values

sserdyuk:
I declared them but didn't assign, so it all compiled but crashed later at run time.


EddyVerbruggen:
Thanks!

PR #528, opened 2017-10-18
(fix): Firebase pod 4.4.0 is now used

Spenstar:
Even though the pod file was changed to use the latest version, the Firebase pod version was reverted back to `4.1.1` after the install script ran. Now the correct pod version will be used.


EddyVerbruggen:
Holy smokes, a new Firebase version *again*! Lol.

This looks good - I'll try to release a new version shortly.

PR #536, opened 2017-11-03
add vars to UploadFileResult

deveix:


EddyVerbruggen:
Hi, @deveix! You're right, that belongs there.

Will you be adding more properties to the TS definition file, or can I merge it as is?

deveix:
I may add more in the future but not now because i only need the name

PR #542, opened 2017-11-07
SDK updates

Spenstar:
I needed to manually change the Android SDK version to 11.4.0 in order to get the build to work any time I re-installed the plugin. Might as well help out and update the main repo. 👍🏼

### DIFF
--- a/publish/scripts/installer.js
+++ b/publish/scripts/installer.js
@@ -215,7 +215,7 @@ function writePodFile(result) {
     }
     try {
         fs.writeFileSync(directories.ios + '/Podfile',
-`pod 'Firebase', '~> 4.4.0'
+`pod 'Firebase', '~> 4.5.0'
 pod 'Firebase/Database'
 pod 'Firebase/Auth'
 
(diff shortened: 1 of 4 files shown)


EddyVerbruggen:
That's very helpful of you, thanks!

Btw, do you have any idea why you need 11.4.0 for "the build to work"? Thx!

Spenstar:
I believe it's because of the `nativescript-geolocation` plugin. It's Google play services version is set to `11.4.0`

EddyVerbruggen:
Ah OK. Yeah those versions are such a pita. I'll merge this soon!

Spenstar:
I hear ya, haha. SDK version mismatches from the different plugins can be a pain.

EddyVerbruggen:
Yeah - to that end we've added a property you can set so all plugins use the same version. Well, only a few plugins I know honor this property really. But Firebase and Geolocation are among them. So if you want, you can try:

https://github.com/NativeScript/nativescript-geolocation#version-conflicts--google-play-services

Spenstar:
Yeah, I remember trying that a while back and still got the same error 🤔 The only way I have been able to get the Android build to be happy is to manually edit the `include.gradle` files found in the plugins.

EddyVerbruggen:
OK. Note that the Geolocation only supports it since 4.2.0 which was released about 2 weeks ago.

Anyhow, I'll merge this soon because I like to keep up with Firebase SDK updates anyway.

EddyVerbruggen:
Yeah I'm not too fond of that ;)

Spenstar:
No worries 😁

EddyVerbruggen:
Latest bump: #590.

PR #598, opened 2018-01-05
Added firestore support for JS Dates and Arrays 

jr76:
Added support for JS Dates and Arrays to the firebase.toValue and firebase.toHashMap functions.


EddyVerbruggen:
Looks good, thank you! 👍

PR #549, opened 2017-11-21
Basic integration of Crashlytics for iOS...

DickSmith:
...including automatic dSYM uploads. Based on @hypery2k 's `nativescript-fabric` plugin.

Since it is only for iOS right now, both original Firebase `crash_reporting` and `crashlytics` can be enabled. If `crashlytics` is enabled, then `crash_reporting` is disabled for iOS, but still enabled for Android.
I prioritized Crashlytics for iOS since the original `crash_reporting` works well enough for Android issues (even with uglify on, but no obfuscation on proguard), but for iOS without dSYM it's almost impossible to decode most of them. The Crashlytics for Firebase offered the easiest solution for automatic dSYM uploading. For Crashlytics dSYM to be properly generated, bitcode must be disabled, which is done via the podfile for all targets.

I have this running in Prod with 30,000+ daily users.

TODOs: implement Android; integrate custom logging

PR #647, opened 2018-03-20
Allow standalone Analytics import

DickSmith:
I wanted to be able to stick all my analytics stuff onto a worker thread, but for iOS was unable to because several of the delegates were being duplicated, even though those delegates aren't directly related to analytics.
This setup is backwards compatible so that `analytics` can either be used off the main `firebase` module, or as a standalone module, with typings being correctly exported for both usages as well.

The same thing could probably be done for some of the other portions of the plugin as well, like `firestore`, etc.

This branch is an extension of my fork from #549, which I assume you'll merge before this one, so those changes are present here as well.


EddyVerbruggen:
Thanks @DickSmith! This is a good starting point for further decoupling. FYI I had to change the export of analytics a little to satisfy my IDE, but it was an easy fix.

PR #648, opened 2018-03-21
Implement setUserId for analytics

olachris:
This adds a "setUserId" method which actually calls https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.html#setUserId(java.lang.String) for Android and https://firebase.google.com/docs/reference/ios/firebaseanalytics/api/reference/Classes/FIRAnalytics#setuserid for IOS

### DIFF
--- a/src/firebase.android.ts
+++ b/src/firebase.android.ts
@@ -550,6 +550,24 @@ firebase.analytics.logEvent = arg => {
   });
 };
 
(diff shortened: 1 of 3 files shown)


EddyVerbruggen:
Your code looks great! I'll add a bit to the readme. Cheers! 👍

Issue #657, opened 2018-03-27
Login with LoginType.PHONE only completes the promise for the first time


jnorkus:
Platform: Android

```
firebase.login({
    type: firebase.LoginType.PHONE,
    phoneOptions: {
        phoneNumber: '+37011'
    }
}).catch(error => {
    console.log(error);
});
```

First time this works as expected, failing and rejecting the promise, `console.log` outputing the error message. Second time the promise is not rejected.

I've debugged this all the way to `firebase.android.js`, this line:
`var OnVerificationStateChangedCallbacks = com.google.firebase.auth.PhoneAuthProvider.OnVerificationStateChangedCallbacks.extend({`

The callbacks work as expected and actually call `reject(errorMessage);` but for some reason after the first failure any subsequent calls will not reject the promise.

I would be willing to help in fixing this (if indeed it is a bug) but I'm pretty new to Nativescript and mobile development and not sure what the `extend()` call is or how to debug native code. It seems to be the root of the problem. I have tried calling the callback directly on the object returned by `extend` and could reproduce.

UPDATE:
Isolated reproduction. Please note that it doesn't matter if it's a `resolve` or `reject` - it only works the first time.
```
reproduce().catch(console.log);

function reproduce() {
    return new Promise((resolve, reject) => {
        var OnVerificationStateChangedCallbacks = com.google.firebase.auth.PhoneAuthProvider.OnVerificationStateChangedCallbacks.extend({
            onVerificationCompleted: (phoneAuthCredential) => {
                resolve(phoneAuthCredential);
            },
            onVerificationFailed: function (firebaseException) {
                reject(firebaseException);
            }
        });

        const cbs = new OnVerificationStateChangedCallbacks();
        cbs.onVerificationFailed('something went wrong');
    });
}
```

jnorkus:
Could be related
https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/462

EddyVerbruggen:
Indeed, looks like #462 is related. This will be fixed with 5.2.0.

jnorkus:
@EddyVerbruggen  
The fix is not complete - `this` is not used in the `onComplete` callback. This results in promise not being resolved after first reject. Strangely though, the next time resolve is called it works.

PR #659, opened 2018-03-28
Fix phone auth login promise resolve/reject #657

jnorkus:
Storing the resolve/reject functions on `this` fixes the issue #657. I'm assuming it has to do with marshalling and this way they get passed by reference as opposed to passing by value. I am happy to update if there is a better and more explicit way to do this.

### DIFF
--- a/src/firebase.android.ts
+++ b/src/firebase.android.ts
@@ -1018,9 +1018,12 @@ function toLoginResult(user) {
 
 firebase.login = arg => {
   return new Promise((resolve, reject) => {
+    this.resolve = resolve;
+    this.reject = reject;
+
     try {
       if (!firebase._isGooglePlayServicesAvailable()) {
-        reject("Google Play services is required for this feature, but not available on this device");
+        this.reject("Google Play services is required for this feature, but not available on this device");
         return;
       }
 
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Wow. This is a nice catch @jnorkus! I guess NativeScript only creates the native class once (one that extends the abstract `OnVerificationStateChangedCallbacks` class), and can't reference the promise the second time because of that.

Also, showing an AdMob banner has a similar issue because that also uses this `.extend({` construct.

EddyVerbruggen:
So I found 2 (other) ways to tackle this, based on our observations:

1. Replace `new OnVerificationStateChangedCallbacks()` with `new OnVerificationStateChangedCallbacks().bind(this)`, or
2. Once the class is extended, cache it.

I'd rather go with the less messy code (1).

EddyVerbruggen:
I'm going to close this PR and THANK YOU very much for pointing this out AND providing a good solution to the problem. Mine esentially does the same but it's much less code. Cheers!

EddyVerbruggen:
Final note: that first option didn't actually work (it always rejected the promise, and that was what I was testing at the time). But I've changed the implementation a bit and now it's working just fine.

Issue #661, opened 2018-03-28
[iOS] FIRDocumentReference.onSnapshot always exists


vcooley:
There is an error in the `FIRDocumentReference.onSnapshot` callback which sets `DocumentSnapshot.exists` to `true` even if the document does not exist. After investigation, it looks like the `DocumentSnapshot.data` method is trying to coerce a null value improperly when invoked.

EddyVerbruggen:
Thanks for the PR!

PR #662, opened 2018-03-28
fix snapshot creation logic and data callback return type

vcooley:
Addresses #661 

Note that this brings the plugin API in line with the web API which could cause breaking changes for those using checks like `DocumentSnapshot.data === null`. If you want to use checks like that, you should use `DocumentSnapshot.data() === undefined` now.


EddyVerbruggen:
Awesome! I’ll review this shortly.

EddyVerbruggen:
That's a great suggestion! The goal is to be as much Web-API-compatible as possible, so I love this change. I'll merge and add your suggestion as well. THANKS 👍

PR #679, opened 2018-04-17
Fix phone authentication promise

jnorkus:
#657


EddyVerbruggen:
It looks like such silly code, but I'm afraid that's how it needs to be 😉
Thanks!

jnorkus:
Yeah I know right, it's far from perfect but I guess that's all we can do with NS right now. Could be an interesting project investigating deeper on how NS does marshalling and how `extend` works but I really don't have time for it 😞

PR #691, opened 2018-04-30
use googlePlayServicesVersion instead of firebase version

breningham:
Now that the firebase and google play services's versions are aligned, it avoids version conflicts with google play services. 

Possible breaking changes: 
 - Using an Old version of Google Play services where there wasnt a comparable firebase version will probably not work and will need updating.

### DIFF
--- a/publish/scripts/installer.js
+++ b/publish/scripts/installer.js
@@ -461,6 +461,7 @@ repositories {
 }
 
 def supportVersion = project.hasProperty("supportVersion") ? project.supportVersion : "26.0.0"
+def googlePlayServicesVersion = project.hasProperty('googlePlayServicesVersion') ? project.googlePlayServicesVersion : "12.0.1"
 
 dependencies {
     compile "com.android.support:appcompat-v7:$supportVersion"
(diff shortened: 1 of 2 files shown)


EddyVerbruggen:
Yeah that's a very good point (about the breaking change). Could be trick if folks want to use the latest Firebbase features (exposed by the plugin) and pin to an old Play Services version that doesn't support the feature.

Perhaps we could issue a big fat warning in the console (triggered by the gradle script) whenever the Play Services version of the project is lower than specified by the plugin?

breningham:
Not sure how i would do that, but i can take a look at it, is there any information about when firebase changed over to googlePlayServices versions?

EddyVerbruggen:
Something like `logger.warn('Please update your Play Services blah blah...')` or (if that doesn't work) `println 'Please update your Play Services blah blah...'` in case the version is [too low](https://stackoverflow.com/a/36400405/2596974)?

publish/scripts/installer.js
```diff
@@ -463,13 +463,18 @@ repositories {
 def supportVersion = project.hasProperty("supportVersion") ? project.supportVersion : "26.0.0"
 def googlePlayServicesVersion = project.hasProperty('googlePlayServicesVersion') ? project.googlePlayServicesVersion : "12.0.1"
 
+if ( VersionNumber.parse( googlePlayServicesVersion ) < VersionNumber.parse( '9.2' ) ) {
+    throw new GradleException('Google Play services version too low, please update to at least 9.2');
```
EddyVerbruggen: ❤️

EddyVerbruggen: I think I'll add also add a warning in case the version is lower than `googlePlayServicesVersion`

breningham:
Hmm, yeah,

I've added the following in the installer.js (ATM)
```(groovy)
if ( VersionNumber.parse( googlePlayServicesVersion ) < VersionNumber.parse( '9.2' ) ) {
    throw new GradleException('Google Play services version too low, please update to at least 9.2');
}
```
using version 9.2 as thats the last version according to https://firebase.google.com/support/release-notes/android 

And i throw an error as to stop it from building as at that point it wont be buildable.

breningham:
Im updating this as during testing i found that you need to be on atleast 11.4.2 for firestore so right now im setting the limit to 11.4.2 instead of 9.2, BUT i will look into making it conditional (as in if the dev wants to use firestore but is on 9.2 then throw the error, but if they aren't then we let them continue)

breningham:
I think this is ready to merge. 

Thoughts?

EddyVerbruggen:
I think so too, thanks a bunch!

PR #716, opened 2018-05-21
Tell user to read the documentation if google-services.json does not exist

Burgov:


EddyVerbruggen:
Good idea, thx!

ibnYusrat:
Might aswell want to mention which part of `documentation` should be read.

PR #718, opened 2018-05-21
expose raw barcode scan result object

Burgov:
I figured it may be nice to expose the raw object until it's fully covered in the plugin's API. For instance I can now use this to retrieve the location of a barcode within the camera and use it to display a label on top of the actual barcode location

### DIFF
--- a/src/mlkit/barcodescanning/index.android.ts
+++ b/src/mlkit/barcodescanning/index.android.ts
@@ -38,7 +38,8 @@ export class MLKitBarcodeScanner extends MLKitBarcodeScannerBase {
           const barcode = barcodes.get(i);
           result.barcodes.push({
             value: barcode.getRawValue(),
-            format: BarcodeFormat[barcode.getFormat()]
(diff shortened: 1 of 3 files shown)


EddyVerbruggen:
Hi, exposing the native object is a pretty common pattern in NativeScript, but the exact implementation is a little different.

Consider an app using the updated TS definition (index.d.ts) that now exposes `FIRVisionBarcode` to the app. Unless the app has the iOS platform typings installed, it will not be able to understand what it is, leading to compilation issues.

The way this is usually handled is adding the `ios` and `android` properties like this:

```typescript
  value: string;
  format: string;
  ios?: any; /* FIRVisionBarcode */
  android?: any; /* whatever.class.this.Is */
```

, and in the implementation files you set either the ios or android properties.

Would you mind updating the PR?

Burgov:
@EddyVerbruggen done

EddyVerbruggen:
Perfect! 👍

PR #719, opened 2018-05-21
trigger handler when there are no results

Burgov:
As I was using this plugin I found that I'd prefer if it also triggered when there are no barcodes on the screen. This looks like different behavior to nativescript-barcodescanner, but it makes sense, because this plugin yields a list, which can be empty.

In my particular case I'm trying to display labels on top of barcodes that are scanned, but because of the way the plugin works, the labels remain even when the barcodes move out of the camera (unless another barcode is visible).

Implementation wise, for the end user of the plugin it's easier to cache the last result and replicate the current behavior, than to build a timeout mechanism to build the described desired behavior.

### DIFF
--- a/src/mlkit/barcodescanning/index.android.ts
+++ b/src/mlkit/barcodescanning/index.android.ts
@@ -23,8 +23,6 @@ export class MLKitBarcodeScanner extends MLKitBarcodeScannerBase {
     return new com.google.android.gms.tasks.OnSuccessListener({
       onSuccess: barcodes => {
 
-        if (barcodes.size() === 0) return;
-
         // const imageSource = new ImageSource();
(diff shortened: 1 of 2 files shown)


EddyVerbruggen:
I agree, but in combination with the changes in #718 (which will send the native object to the app through the notification system) I'm afraid it may crash your app.

Perhaps you can remove the labels in case the barcode is not detected since x milliseconds?

Burgov:
How would that crash the app? If there are no results there's also no raw results outputted, or am I missing sth?

Your suggestion would work indeed. I figured the developer would be helped more with an empty list, but either way will work.

EddyVerbruggen:
You're right, I was confused with returning the actual image that was used to scan for barcodes, but that's not what that PR was about. I'll consider including this for 6.1.0.

PR #721, opened 2018-05-21
also add message for iOS

Burgov:
I added another commit to handle iOS but you had already merged the PR :) Here it is


EddyVerbruggen:
😎❤️

PR #726, opened 2018-05-25
Fix for issue #488 - prevent cyclic notification message structure on iOS

abhayastudios:
Prevent cyclic notification message structure on iOS


abhayastudios:
@EddyVerbruggen hmm something went wrong here on the Android webpack build, but my (small) change is only on `firebase.ios.ts`...

EddyVerbruggen:
Looks good, thanks! That Travis Android issue is indeed not something you'll need to worry about.

PR #741, opened 2018-06-04
fix(ui): temp fix for app becoming unresponsive due to infinite animation

shiv19:
App keeps trying to play the swing keyframe animation even when the view is destroyed, ending up in a play-catch loop.

to avoid this, animation iteration count was fixed to 10.


EddyVerbruggen:
Cheers mate!

Issue #730, opened 2018-05-30
iOS app crashes when opening push notification when app is killed


davidsandoz:
As I faced this issue and looked for a solution, I found that it has already been fixed in the past with #520.

However, the fix brought in that PR was changed in c6bb8a7829662caca7d2c9645441f68f2547d310 and 3e5a2ea07a0571f402850169579f976597832794 because of #564.

Now that `FIRApp.configure()` is in `applicationDidFinishLaunchingWithOptions()`, when `FIRAuth.auth().canHandleNotification(notification)` [is called](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/src/firebase.ios.ts#L60) in `applicationDidReceiveRemoteNotificationFetchCompletionHandler()`, then the app crashes.

I simply tried to comment the whole `if` block where `FIRAuth` is used to see if it worked better and indeed, the app isn't crashing anymore. However, if I test that `FIRAuth` is not undefined, then the app still crashes.

The iOS app Crash log also confirms that the error is coming from that line:
```
Last Exception Backtrace:
0   CoreFoundation                	0x183667d38 __exceptionPreprocess + 124
1   libobjc.A.dylib               	0x182b7c528 objc_exception_throw + 55
2   CoreFoundation                	0x183667c80 +[NSException raise:format:] + 115
3   nativescript                  	0x102941af0 +[FIRAuth auth] + 95
```

So I suppose this happens because `FIRApp.configure()` hasn't been called yet when it tries to call `FIRAuth.auth()`.

But now I don't know where it would be better to have this `FIRApp.configure()`. I saw that it must be called only once, so we cannot copy it in `applicationDidReceiveRemoteNotificationFetchCompletionHandler()`.

Any idea on how to make that work?

PR #745, opened 2018-06-05
Check that firebase is configured before calling FIRAuth.auth()

davidsandoz:
For my use case in EddyVerbruggen/nativescript-plugin-firebase#730, it now works correctly with this fix and I didn't notice any other issues.

But I don't think that I ever need to get into that `if` block.
@EddyVerbruggen Let me know if you think anything needs to be changed.

Closes EddyVerbruggen/nativescript-plugin-firebase#730


EddyVerbruggen:
Crashing is bad. Thank you!

davidsandoz:
Thanks for the merge @EddyVerbruggen! Do you happen to have an ETA for the next release?

EddyVerbruggen:
This week.

Issue #737, opened 2018-06-04
ml kit barcode 


lucasp18:
Hello I would like to thank for developing the firebase plugin and ml barcode.
My English is not so good so I'm writing this text with the help of google translate.
I'm doing some testing with your ml barcode plugin and found some improvements, I'll list it down to what I found:

1- Whenever I open the screen to read the barcode, the camera permission screen appears. I tried some approaches, but none avoided the permission display. Could you improve this or help me improve?

2- The onBarcodeScanResult method is always called twice, would it be possible to do the other plugin, when I read the first code does not read but no code?

3- Could you explain how the plugin reading barcode is configured?
Currently the plugin looks like this:

<MLKitBarcodeScanner
      width = "250"
      height = "250"
      formats = "EAN_13"
      processEveryNthFrame = "5"
      (scanResult) = "onBarcodeScanResult ($ event)">
  </ MLKitBarcodeScanner>

Thank you for the contact

lucasp18:
open PR for resolution item 1

EddyVerbruggen:
Hi,

I don't see the issue with 1. really. The first time (after the app is installed) the permission popup will appear when you access the camera, but the second time it will no longer request the permission. Can you share a video where you see this permission popup every time?

Note that requesting permission is not supposed to show the permission prompt in case permission has already been granted.

PR #752, opened 2018-06-08
Add checkPermission for camera before requestPermission

lucasp18:
Hello.

I created this problem (https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/737) and found a way to solve the permissions problem. I just changed the .ts file, I do not know if I need to change the .js file as well
I hope I have solved the problem correctly. Please, disregard some mistake this is my first PR

### DIFF
--- a/src/mlkit/mlkit-cameraview.android.ts
+++ b/src/mlkit/mlkit-cameraview.android.ts
@@ -55,31 +55,36 @@ export abstract class MLKitCameraView extends MLKitCameraViewBase {
   createNativeView(): Object {
     let v = super.createNativeView();
 
-    if (this.hasCamera()) {
-      const permissionCb = (args: application.AndroidActivityRequestPermissionsEventData) => {
-        if (args.requestCode === CAMERA_PERMISSION_REQUEST_CODE) {
-          application.android.off(application.AndroidApplication.activityRequestPermissionsEvent, permissionCb);
+    if (this.hasCamera()) {      
+      if (android.support.v4.content.ContextCompat.checkSelfPermission(application.android.foregroundActivity || application.android.startActivity, android.Manifest.permission.CAMERA) != android.content.pm.PackageManager.PERMISSION_GRANTED) {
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Please see my comment https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/737#issuecomment-398361643

EddyVerbruggen:
@lucasp18 Can you please check out that comment?

dmdnkv:
Hello.
@EddyVerbruggen, the problem is really there. Just experienced it recently with nativescript-plugin-firebase@9.0.0 and Android 6 (API 23). The fix from @lucasp18 helped. Here we have a different behavior of ActivityCompatClass.requestPermissions on Android 6 and Android 8.1. It asks for the permission every time whenever a user opens a screen with MLKitBarcodeScanner.

AndroidManifest.xml has the following users-permission
```	
<uses-permission android:name="android.permission.INTERNET"/>
<uses-permission android:name="android.permission.CAMERA"/>
```

app.gradle is configured with the following sdk options
```
android {  
  defaultConfig {
    minSdkVersion 23
    targetSdkVersion 28
```

The angular component that holds the second screen has the following markup:
```
<header>
    <Button (tap)="goBack()" text="go back"></Button>
</header>
<MLKitBarcodeScanner></MLKitBarcodeScanner>
```

goBack performs just common RouterExtensions call
```
goBack() {
  this.routerExteions.back();
}
```

MLKitBarcodeScanner on Android 6 (API 23)
ns_plugin_firebase_900_android6_api23
MLKitBarcodeScanner on Android 8.1 (API 27)
ns_plugin_firebase_900_android81_api27

EddyVerbruggen:
@dmdnkv Thanks for those video's? Why the beeeeeep does Android prompt users another time once permission has been granted? Smells like a bug in their SDK to me (that we have to work around).

EddyVerbruggen:
By the way, I just confirmed this behavior on an Android emulator (API 23).

PR #779, opened 2018-07-02
iOS GeoPoint fix: read Geopoint from correct library

vcooley:
iOS is currently using the declaration file to attempt to create a geopoint. This fixes the iOS file to read from the firebase common file instead.


EddyVerbruggen:
Thanks a bunch! 💯

PR #794, opened 2018-07-10
enable querying by GeoPoint

vcooley:
Currently, GeoPoints are saved and read correctly, but we're lacking query functionality. This PR implement such functionality.


EddyVerbruggen:
Very cooley! This looks good, I'll merge it right now and test it when releasing a new version (not sure when that'll be yet though).

PR #814, opened 2018-07-24
allow updateProfile arguments if they are displayName or photoURL. Do…

ThomCM:
… not require both.


EddyVerbruggen:
This would do the exact opposite of what you describe.. `(!a || !b)` means both `a` and `b` will become required. AFAIK one of these values are required (but not both).

Issue #855, opened 2018-08-17
iOS: TypeError: null is not an object (evaluating 'snapshot.documentID')


jnorkus:
After logging out iOS app crashes with the error in title. Seems related to the issue where it is not possible to unsubscribe properly.

EddyVerbruggen:
I've added a guard to the only place that error could originate from. Thanks for reporting it!

jnorkus:
I had basically done the same thing in the PR that references this issue. But I guess you can close it now also.

EddyVerbruggen:
I feel bad for not seeing it. Sorry man!

PR #856, opened 2018-08-17
fix: prevent iOS from crashing on logout

jnorkus:


jnorkus:
https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/855

EddyVerbruggen:
Damn, totally missed this one, sorry!

PR #869, opened 2018-08-24
[DO NOT MERGE] Tbozhikov/interactive push

tbozhikov:
This PR implements functionality that is present in push-plugin and not present here (firebase), relates to the eventual [push-plugin retirement](https://github.com/NativeScript/push-plugin/issues/230). What is done:
- extract messaging functionality in separate modules for iOS an Android
- add interactive push notifications support for iOS
- interactive push notifications example in demo (TODO in demo-ng) 
- add **areNotificationsEnabled()** API for getting if user has enabled notifications
- **areNotificationsEnabled()** API example in demo (TODO in demo-ng)
- Clear badges only when app is active. This fixes a behavior when a silent push is sent and badge is set to a value. In this case the plugin used to clear the badge after setting it to the desired value

Still work in progress, comments/reviews are welcome!


demo/app/messaging-view-model.ts
```diff
@@ -0,0 +1,165 @@
+import { alert } from "tns-core-modules/ui/dialogs";
+import { isIOS } from "tns-core-modules/platform";
+import { ios as iOSUtils } from "tns-core-modules/utils/utils";
+import * as firebase from "nativescript-plugin-firebase";
+import { IOSPushSettings, IosInteractiveNotificationActionOptions, IosInteractivePushSettings, PushNotificationModel } from "nativescript-plugin-firebase/messaging/messaging";
```
EddyVerbruggen: Perhaps it's nicer to use the same casing for all these objects: `IOS..` vs `Ios..` looks a bit sloppy 😉

src/firebase.ios.ts
```diff
@@ -2216,9 +1804,13 @@ firebase.firestore.collection = (collectionPath: string): firestore.CollectionRe
 
 firebase.firestore.onDocumentSnapshot = (docRef: FIRDocumentReference, callback: (doc: DocumentSnapshot) => void): () => void => {
   const listener = docRef.addSnapshotListener((snapshot: FIRDocumentSnapshot, error: NSError) => {
+<<<<<<< HEAD
```
EddyVerbruggen: Woops, there's an unresolved merge conflict in here 😇

src/firebase.android.ts
```diff
@@ -1,4 +1,5 @@
 import { DocumentSnapshot, firebase, GeoPoint, QuerySnapshot } from "./firebase-common";
+import * as firebaseMessaging from "./messaging/messaging";
```
EddyVerbruggen: I'm getting `firebase.android.ts:2:36 - error TS2307: Cannot find module './messaging/messaging'.`, I think there's an `index.d.ts` missing in Git. Probably because my rules in the `.gitignore` file are a bit silly (I'm ignoring those files in there without a wildcard).

EddyVerbruggen:
GREAT work, thanks a lot!

PR #893, opened 2018-09-05
[Android] MLKit camera resume when app resumes from background.

Soarc:
When live camera feed is running and application is moved to the background, on resume camera is not turned on. This pull request fix it.


Soarc:
Looks like npm run package is missing from travis build definition. @EddyVerbruggen should i fix it?

EddyVerbruggen:
Works perfectly fine for me, thanks!

I'll try to curt the 6.8.0 release asap.

PR #899, opened 2018-09-07
[iOS] Fix special fields before adding documents

vcooley:
Currently the add function on Firestore skips the step where it converts specific field types to the native instances, causing incorrect data to be saved to the database.


EddyVerbruggen:
Looks good to me. Thanks!

PR #901, opened 2018-09-10
Improved docs for google authentication

vratojr:
To was not clear at all the phrase: 

 add the _Web SDK configuration_.

And it took a while to figure out that in the end was to update my already existing google service file :)
So maybe this could be misleading for others.

Also, fixed some broken link.


EddyVerbruggen:
Thank you!

Issue #894, opened 2018-09-05
Incorrect Google Play Services version check in include.gradle


Danziger:
In the file `include.gradle` there's a check:

    if ( VersionNumber.parse( googlePlayServicesVersion ) < VersionNumber.parse( '15.0.0' ) ) {
        throw new GradleException(" googlePlayServicesVersion set too low, please update to at least 15.0.0 ( currently set to $googlePlayServicesVersion )");
    }

Which is wrong, as it will report a version of `15.0.+` as being too low as well. It should rather be:

    if ( VersionNumber.parse( googlePlayServicesVersion ) < VersionNumber.parse( '15.0.+' ) ) {
        throw new GradleException(" googlePlayServicesVersion set too low, please update to at least 15.0.0 / 15.0.+ ( currently set to $googlePlayServicesVersion )");
    }

I would open a PR myself instead of an Issue, but I can't find that file as it looks like it's being generated. Related to this, there's a broken link to this file in the README, under `Google Play Services Version`:

> The plugin will default to [this version](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/src/platforms/android/include.gradle#L27) of the Android `play-services-base`...

DickSmith:
@Danziger 
https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/src/scripts/postinstall.js
That is the file that generates these, I believe.

However, to get around it for now you can also do this in your `app.gradle`:
```
dependencies {
    configurations.all {
        exclude group: 'commons-logging', module: 'commons-logging'
        resolutionStrategy.eachDependency { DependencyResolveDetails details ->
            def requested = details.requested
            if (requested.group == 'com.google.android.gms' || requested.group == 'com.google.firebase') {
                details.useVersion '+'
            } else if (requested.group == 'com.android.support' && requested.name != 'multidex') {
                // com.android.support major version should match buildToolsVersion
                details.useVersion '27.+'
            }
        }
    }
    implementation "com.google.android.gms:play-services-tagmanager:15.0.0"
}

project.ext {
  // TODO PR this so that '15.+' can be used
    googlePlayServicesVersion = "15.0.0"
    supportVersion = "27.+"
}
```
As you can see there, I've also been meaning to PR this, but it needs to be able to handle the `+` in any of the 3 positions, as well.

[...]

Issue #882, opened 2018-08-28
googlePlayServicesVersion problem [with temporary solution, should i do PR?]


indradha:
Found this bug, maybe just me or some other with the googlePlayServicesVersion problem, so am sharing my solution/fix here that takes me hours to solve:

As stated on the documentation, we have to set googlePlayServicesVersion to 15.0.0, but somehow I got this error

```
Failed to capture snapshot of input files for task ':app:preDebugBuild' property 'compileManifests' during up-to-date check.
The library com.google.android.gms:play-services-auth-base is being requested by various other libraries at [[15.0.0,15.0.0]], but resolves to 15.1.0. Disable the plugin and check your dependencies tree using ./gradlew :app:dependencies.

```

without setting the googlePlayServicesVersion, my error will fall on 15.0.1:

`The library com.google.android.gms:play-services-auth-base is being requested by various other libraries at [[15.0.1,15.0.1]], but resolves to 15.1.0. Disable the plugin and check your dependencies tree using ./gradlew :app:dependencies.`

The temporary fix/solution in my code will be updating include.gradle (is it set by scripts/installer.js?) on these lines below 

```
compile "com.google.android.gms:play-services-base:$googlePlayServicesVersion" -> compile "com.google.android.gms:play-services-base:15.0.1"

compile "com.google.android.gms:play-services-auth:$googlePlayServicesVersion" -> compile "com.google.android.gms:play-services-auth:16.0.0"

``` 

(I got the latest version from here: https://developers.google.com/android/guides/setup)

sebastiandg7:
Same happening here

andermev:
The same for me, I looked up this issue in several pages, until finally I achieved resolve it.

I am going to share you my **app.gradle**

```
dependencies {
  compile "com.google.android.gms:play-services-auth:16.0.0"
  compile fileTree(dir: 'libs', include: ['*.jar'])
}

android {  
  defaultConfig {  
    generatedDensities = []
    multiDexEnabled true
    applicationId = "org.nativescript.appmigotabsapp"
  }  
  aaptOptions {  
    additionalParameters "--no-version-vectors"  
  }

  project.ext {
    googlePlayServicesVersion = "15.0.0"
  }

  buildTypes {
      release {
          minifyEnabled false
          proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
      }
  } 

  dexOptions {
    jumboMode true
[...]

PR #908, opened 2018-09-14
googlePlayServicesVersion bugfix [with temporary solution]

indradha:
based on issue #882, possibly fix for issue #894


EddyVerbruggen:
Thanks! I reverted the first change, but the auth change looks good to me.

PR #912, opened 2018-09-17
Allows disabling automatic notification creation and always showing notifications when in foreground

Danziger:
This should fix conflicts when using this plugin together with `nativescript-local-notifications`.

Fixes https://github.com/EddyVerbruggen/nativescript-local-notifications/issues/107#issuecomment-419902485


EddyVerbruggen:
Impressive work (again), thanks!

PR #914, opened 2018-09-18
Adds missing options (they two new ones I added) to the InitOptions interface

Danziger:
@EddyVerbruggen Just forgot about this.


EddyVerbruggen:
That’s pretty useful, thx!

Issue #903, opened 2018-09-11
user.additionalUserInfo are (almost) never returned as stated in the docs


vratojr:
Hello @EddyVerbruggen,
In the [docs](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/docs/AUTHENTICATION.md#whats-returned-when-logged-in) there's written:

    All login functions below, as well as getCurrentUser return a 'User' object with these properties:

The fact is that it's not doing it :)
For example, I checked the code and I see that _getCurrentUser_ is calling _toLoginResult_ without the additional parameter required to fill the extra data.

This is also true for other situations (like for example when listening for state change). So it's suspect :)
Is there a problem (that I don't see) in always returning the additional info (and maybe the docs are outdated and we should remove that statement) or simply we can modify the code to always return the additionaldata?

I cannot test it sorry as I still unable to build this plugin and testing it locally... :/

EddyVerbruggen:
Do these instructions work for you? https://github.com/EddyVerbruggen/nativescript-plugin-firebase#demo-app

vratojr:
Hello Eddy!
You mean running the demo and testing the behaviour of getCurrentUser?
I confirm that for the native api (sorry I should have mentioned it before)
getCurrentUser doesn't return the additionalUserInfo.

vratojr:
Hello Eddy!
You mean running the demo and testing the behaviour of getCurrentUser?
I confirm that, for the **native api** (sorry I should have mentioned it before), getCurrentUser doesn't return the additionalUserInfo.

PR #919, opened 2018-09-24
Storing additional user info as local state

vratojr:
Hello Eddy, since I didn't have any news from your side, I tryed to solve the problem my way :)

As stated in #903, additionalUserInfo are not retrieved when calling getCurrentUser using nativeAPI.
So in this approach I try to mimick what is being done for the web api, that is, storing the additional info as a global object and returning it when asked for.

Android only.


EddyVerbruggen:
Interesting approach, thank you!

EddyVerbruggen:
Added a similar implementation for iOS.

Will be available in 7.1.0 (released later this week).

Issue #947, opened 2018-10-07
Nested GeoPoint gets saved as object in Firestore on iOS


abhayastudios:
GeoPoint gets saved as object in Firestore on iOS with plugin `7.1.2` (and `6.8.1` for that matter). On Android it works fine.

EddyVerbruggen:
Hi, can you share the call you use?

abhayastudios:
@EddyVerbruggen let me know if you were able to reproduce or whether you need me to create a little app to demonstrate the issue

EddyVerbruggen:
That's strange.. I just tested this in the demo-ng app (iOS sim, latest plugin version):

```typescript
import { firestore } from "nativescript-plugin-firebase";
const firebase = require("nativescript-plugin-firebase/app");

  lastKnownLocation: firebase.firestore().GeoPoint(4.34, 5.67),
  lastKnownLocation2: firestore.GeoPoint(4.34, 5.67)
```

Result:

<img width="553" alt="screenshot 2018-10-08 at 17 28 06" src="https://user-images.githubusercontent.com/1426370/46618353-8cb69200-cb1f-11e8-8f31-a7da27c7dc8c.png">

I will try `.set` as well.

EddyVerbruggen:
Hmm, `.set` works equally well for me with similar code.

abhayastudios:
Thanks for checking @EddyVerbruggen. I will try to replicate in a minimal project.

abhayastudios:
Hi @EddyVerbruggen,

So it seems that the issue occurs only when using GeoPoint nested in an object, such as below.

```
let profile:any = {
      first: 'First',
      last: 'Last',
      location: {
        geoPoint: firestore.GeoPoint(52.345342, 4.900721400000066),
        latitude: 52.345342,
        longitude: 4.900721400000066
      }
    }

    let geoRef:firestore.CollectionReference = firestore.collection("geopoints");

    geoRef.doc('issue947').set(profile)
    .then(() => { console.log("Saved GeoLocation data to Firestore!"); })
    .catch((error) => console.log(`Error saving GeoLocation data to Firestore: ${error}`));
  }
```

Is this expected behavior? As said, on Android this still produces a GeoPoint type in Firestore.

Thanks!

EddyVerbruggen:
That could indeed be a limitation, but let's see if it's fixable. Thanks for investigating this!

EddyVerbruggen:
[...]

PR #952, opened 2018-10-09
Fix on fix for issue #947

abhayastudios:
Following up on #951, this seems to work better with Date and other data types as well.


EddyVerbruggen:
Let me merge and test this, thanks again!

PR #960, opened 2018-10-14
EmailLinkAuthentication - Don't check action type when resuming app.

vratojr:
Hello Eddy,
Email link authentication to me is not working in android.
That's because at startup there is a check on the action type:

        const isLaunchIntent = "android.intent.action.VIEW" === intent.getAction();

That seems to be always false because in reality the type of intent is: _android.intent.action.MAIN_
So, I propose to drop the check on the action type (even in firebase examples they are not checking it) and rather rely on the following check:

          if (firebaseAuth.isSignInWithEmailLink(emailLink)) {

that should prevent us to enter auth code when we should not.

What do you think?

Cheers

### DIFF
--- a/src/firebase.android.ts
+++ b/src/firebase.android.ts
@@ -47,13 +47,10 @@ const dynamicLinksEnabled = lazy(() => typeof (com.google.firebase.dynamiclinks)
       firebaseMessaging.onAppModuleLaunchEvent(args);
     }
 
-    const intent = args.android;
-    const isLaunchIntent = "android.intent.action.VIEW" === intent.getAction();
-
-    if (isLaunchIntent && dynamicLinksEnabled()) {
+    if (dynamicLinksEnabled()) {
       // let's see if this is part of an email-link authentication flow
       const firebaseAuth = com.google.firebase.auth.FirebaseAuth.getInstance();
-      const emailLink = "" + intent.getData();
+      const emailLink = "" + args.android.getData();
       if (firebaseAuth.isSignInWithEmailLink(emailLink)) {
         const rememberedEmail = firebase.getRememberedEmailForEmailLinkLogin();
         if (rememberedEmail !== undefined) {


EddyVerbruggen:
Your change is actually very similar to what it used to be, but in some case I can't recall it caused an issue. Let's merge this and test things a bit. Thanks!

mmeven:
Hello,

For some reason the plugin does not seem to work on android after this change, when authentication in firebase.nativescript.json is set to false. I think it's related to your modification because the crash happens on this line :

`var firebaseAuth = com.google.firebase.auth.FirebaseAuth.getInstance();`

The error is : 

`TypeError: Cannot read property 'getInstance' of undefined`
`File: "file:///data/data/com.cadreo.android.dev/files/app/tns_modules/nativescript-plugin-firebase/firebase.js, line: 39, column: 69`

Maybe it misses a check to see if the authentication function is enabled. Enabling it fixes the error.

mmeven:
Hello,

Your modification somehow put another problem in evidence. The code within this condition :

`if (dynamicLinksEnabled()) {`

is now executed more often, which was intended if I understood well. But it misses a check to see if 'authentication' is set to true in the firebase.nativescript.json file. If it is set to false, the app crashes at the following line, when trying to get the FirebaseAuth instance : 

`var firebaseAuth = com.google.firebase.auth.FirebaseAuth.getInstance();`

This is the corresponding error :

`TypeError: Cannot read property 'getInstance' of undefined`
`File: "file:///data/data/com.cadreo.android.dev/files/app/tns_modules/nativescript-plugin-firebase/firebase.js, line: 39, column: 69`

This seems to be fixed by changing the following check :

`if (firebaseAuth.isSignInWithEmailLink(emailLink)) {`

To something like : 

`if (authEnabled && com.google.firebase.auth.FirebaseAuth.getInstance().isSignInWithEmailLink(emailLink)) {`

I hope that helps :)

EddyVerbruggen:
@mmeven you're absolutely right - your proposed changes will be part of 7.3.0.

PR #969, opened 2018-10-15
Only copy googleservices json/plist if needed

### LINKED ISSUE #968: Constantly rebuilding when running with .dev and .prod separation
Nativescript detects the copy from GoogleService-Info.plist.dev to GoogleService-Info.plist as a file change, even if both files are the same. As such, running `tns run ios` results in endless xcode builds and app installs. Maybe there should be a check to see if the files are the same?

How to reproduce:
Use `GoogleService-Info.plist.dev` and ` GoogleService-Info.plist.prod`
Run `tns run ios`
Check the console for constant rebuilding

edusperoni:
Fixes #968 

Breaking change: keeps old `GoogleService-Info.plist` in App_Resources. This file is always replaced by the produciton/development version if used, so you should add it to .gitignore.

### DIFF
--- a/publish/scripts/installer.js
+++ b/publish/scripts/installer.js
@@ -283,8 +283,8 @@ function echoAndroidManifestChanges(result) {
       console.log('\n######################################################################################################');
       console.log('Open your app\'s resources/Android/AndroidManifest.xml file and add this (see the demo for an example):');
       console.log('<meta-data\n' +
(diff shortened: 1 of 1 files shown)


edusperoni:
closing in favor of #974

EddyVerbruggen:
Thanks!

PR #973, opened 2018-10-16
Add Crashlytics log function

pboulch:


EddyVerbruggen:
Thanks, works as expected

PR #987, opened 2018-10-23
Adding authListener to the list of listeners if passed in 'init'

vratojr:
When the authStateListener is passed through 'init' it's not added to the list of listeners.
This implies that, for example, for an emailLink login, when we have:

    firebase.notifyAuthStateListeners

The state listener is not called.


EddyVerbruggen:
Thanks, once again!

PR #995, opened 2018-10-27
Api to unlink a provider

vratojr:
**The IOS code is just a guess. :)**

Also fixes the build of the demo package.

### DIFF
--- a/src/firebase.android.ts
+++ b/src/firebase.android.ts
@@ -312,7 +312,7 @@ firebase.init = arg => {
         const firebaseAuth = com.google.firebase.auth.FirebaseAuth.getInstance();
 
         if (arg.onAuthStateChanged) {
-          firebase.addAuthStateListener(arg.onAuthStateChanged)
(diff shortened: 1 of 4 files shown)


src/firebase.android.ts
```diff
@@ -663,6 +663,30 @@ firebase.logout = arg => {
   });
 };
 
+firebase.unlink = providerId => {
+  return new Promise((resolve, reject) => {
+    try {
+      const user = com.google.firebase.auth.FirebaseAuth.getInstance().getCurrentUser();
+      if (!user) {
+        resolve();
```
EddyVerbruggen: Note to self: add a `return`, because `resolve` doesn't stop execution.

EddyVerbruggen:
Thanks a lot for this! I'll add Web API compatibility and release it today with plugin version 7.6.0.

EddyVerbruggen:
By the way, you "guessed" iOS implementation works perfectly fine. My advice: guess more often! 🙃

PR #996, opened 2018-10-27
Fixed the way to get dynamic link after first installation

shabib3:
Hi @EddyVerbruggen,

Has I mentioned (a long time ago) here - https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/819 .

There is an issue with getting the dynamic link after installation via the link.
After some digging I found the suggested way to extract it [here](https://firebase.google.com/docs/dynamic-links/android/receive).

After testing it I confirmed it indeed works! :)


EddyVerbruggen:
Hmm, changing `oncomplete` to `onsuccess` makes sense. Thanks a lot!

PR #1017, opened 2018-11-08
Fix for when googlePlayServicesVersion is '+'

edusperoni:
The app build will fail when using googlePlayServicesVersion, if it's set to '+' despite version being >15.0.0


EddyVerbruggen:
Thanks!

PR #1025, opened 2018-11-11
Minimum ios version

nickykln:
Hi @EddyVerbruggen, 
The work you've made on this project, is simply amazing.  Thanks a lot.
I am a new a newbee in nativescript, but I have struggle with a compilation issue.
If you think this additional line can be useful. otherwise, no problem and thanks again for your great job.


EddyVerbruggen:
Oh interesting, thanks! It may be specific to one of the features (MLKit comes to mind). But I can’t imagine folks need to support iOS 8 anymore these days anyway.

PR #1046, opened 2018-11-26
Set 'foreground: false' for iOS Background Notification

cerireyhan:
foreground was always true even if the app is opened from a background notification in iOS.

### DIFF
--- a/src/messaging/messaging.ios.ts
+++ b/src/messaging/messaging.ios.ts
@@ -427,7 +427,13 @@ function _registerForRemoteNotifications() {
           }
         }
 
-        userInfoJSON.foreground = true;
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Hi, thanks!

Was this in case of an interactive notification?

cerireyhan:
No problem :)

It was a regular notification with data payload. Not an interactive notification.

EddyVerbruggen:
Hmm.. according to my own comment this code should only be invoked when in the foreground, but I guess that's inaccurate then. Your change should not hurt. Thanks!

cerireyhan:
Yes, maybe we should change the comment as well to avoid future confusion. Thanks for fast PR check.

Issue #1048, opened 2018-11-28
Android: Remote Config not fetched if called before foregroundActivity is set


jnorkus:
I am upgrading my apps to latest Angular and NS. I've run into this issue where my remote config is not fetched if called before `foregroundActivity` is set. I found this code:
```
// if this is called before application.start() wait for the event to fire
appModule.on(appModule.launchEvent, runGetRemoteConfig);
```
The problem is that launch has already happened at this point and is not fired at all. Looks like `launchEvent` does not guarantee the `foregroundActivity` to be present.

jnorkus:
I created a PR with a fix.

EddyVerbruggen:
TY!

PR #1049, opened 2018-11-28
fix: remote config foregroundActivity instance

jnorkus:
https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/1048


EddyVerbruggen:
Thanks! In another project, I noticed I had to do something similar. I'm going to add some more changes around this, but this is perfect in its own right. 👍

PR #1050, opened 2018-11-29
fix(remote-config): resume event subscription not terminated

jnorkus:
New day fresh eyes, caught this bug from yesterday's PR.


EddyVerbruggen:
Thx. I did something similar (see my change), but this may be neater. Can you remove my ‘off’ handler?

EddyVerbruggen:
Super! 👍

PR #1051, opened 2018-11-29
onAuthStateChangedListener required for AF2 support

larssn:
This PR add native support for `onAuthStateChanged` (named: onAuthStateChangedListener as onAuthStateChanged is in use).

The listener is supposed to emit immediately when registered, as stated here: [Docs](https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuth.html#addAuthStateListener(com.google.firebase.auth.FirebaseAuth.AuthStateListener)). This makes it behave more like the Web SDK, and is required for my AngularFire2 conversion.

Since this is new code, I’ve also taken steps to move towards the Web SDK user object. It doesn’t return a complete web user object yet (and probably never will); I’ve made a TODO to decide what additional methods are required. I figure this is OK, since its not a publicly supported API yet.

An additional side-effect of the above requirement is importing the typings **ONLY** from the official firebase npm module. To only import the typings I use typescript 2.9’s “import” syntax, which prompted an update of typescript (which I brought up to the latest version). Also update tns-core-modules and declarations while I was at it.

I haven’t modified any existing code, so backwards compatibility should be fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ src/platforms/android/include.gradle
 !src/performance/performance.d.ts
 !src/storage/storage.d.ts
 !src/admob/admob.d.ts
+!src/auth/auth.d.ts
 !src/messaging/messaging.d.ts
 !src/mlkit/*/index.d.ts
 !src/mlkit/mlkit-cameraview.d.ts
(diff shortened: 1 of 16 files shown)


larssn:
I'll rebase and redo the PR.

PR #1052, opened 2018-11-30
onAuthStateChangedListener required for AF2 support

larssn:
This PR add native support for `onAuthStateChanged` (named: onAuthStateChangedListener as onAuthStateChanged is in use).

The listener is supposed to emit immediately when registered, as stated here: [Docs](https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuth.html#addAuthStateListener(com.google.firebase.auth.FirebaseAuth.AuthStateListener)). This makes it behave more like the Web SDK, and is required for my AngularFire2 conversion.

Since this is new code, I’ve also taken steps to move towards the Web SDK user object. It doesn’t return a complete web user object yet (and probably never will); I’ve made a TODO to decide what additional methods are required. I figure this is OK, since its not a publicly supported API yet.

An additional side-effect of the above requirement is importing the typings **ONLY** from the official firebase npm module. To only import the typings I use typescript 2.9’s “import” syntax, which prompted an update of typescript (which I brought up to the latest version). Also updated tns-core-modules and declarations while I was at it.

I haven’t modified any existing code, so backwards compatibility should be fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ src/platforms/android/include.gradle
 !src/performance/performance.d.ts
 !src/storage/storage.d.ts
 !src/admob/admob.d.ts
+!src/auth/auth.d.ts
 !src/messaging/messaging.d.ts
 !src/mlkit/*/index.d.ts
 !src/mlkit/mlkit-cameraview.d.ts
--- a/src/app/auth/index.ts
+++ b/src/app/auth/index.ts
@@ -1,7 +1,7 @@
 import * as firebase from "../../firebase";
 import { FirebaseEmailLinkActionCodeSettings, LoginType, User } from "../../firebase";
 
-export module auth {
+export namespace auth {
   export class Auth {
     private authStateChangedHandler;
     public currentUser: User;
@@ -9,7 +9,22 @@ export module auth {
     public onAuthStateChanged(handler: (user: User) => void): void {
       this.authStateChangedHandler = handler;
       console.log(">> added onAuthStateChanged handler");
-    };
+    }
+
+    public onAuthStateChangedListener(
+      nextOrObserver:
+        | import('firebase').Observer<any>
+        | ((a: import('firebase').User | null) => any),
+      error?: ((a: import('firebase').auth.Error) => any),
+      completed?: import('firebase').Unsubscribe,
+    ): import('firebase').Unsubscribe {
+      const listener = (<any>firebase).auth.addAuthStateListener(nextOrObserver, error);
+      return () => {
+        if (listener) {
+          (<any>firebase).auth.removeAuthStateListener(listener);
+        }
+      };
+    }
 
     public signOut(): Promise<any> {
       return new Promise((resolve, reject) => {
@@ -113,7 +128,7 @@ export module auth {
           this.currentUser = user;
           resolve(user);
         }).catch(err => reject(err));
-      })
+      });
     }
 
     public signInAnonymously(): Promise<any> {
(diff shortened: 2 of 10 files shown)


larssn:
Maybe we can discuss renaming `onAuthStateChangedListener` to `onAuthStateChanged`, and flag it as a breaking change for v8, to better align it with the web api.

EddyVerbruggen:
Hi! I'm currently reviewing and testing this PR. I'm hoping we can remove the dependency on `firebase` actually. And I think we can get away with renaming the `onAuthStateChangedListener` to `onAuthStateChanged`. Perhaps we can add a few properties of the old `onAuthStateChanged` return object to the new one for backward compatibility.

larssn:
Well at it stands now, I need the firebase dependency in order to compile, as I import the types from that library. But that is all that's being used, so the bundle size should not increase, however I haven't tested that claim.

> Perhaps we can add a few properties of the old onAuthStateChanged return object to the new one for backward compatibility.

Sure, it's not a problem adding some of the old properties to the new object, and then just flag them as deprecated. They can then be removed later.

Ps. I just pushed a small fix (forgot to check for nulls).

EddyVerbruggen:
> Well at it stands now, I need the firebase dependency in order to compile, as I import the types from that library. But that is all that's being used, so the bundle size should not increase, however I haven't tested that claim.

I was going to check that as well. If it doesn't matter for the end-product then I don't mind at all. Perhaps it could be a "dev dependency" even(?)

> Sure, it's not a problem adding some of the old properties to the new object, and then just flag them as deprecated. They can then be removed later.

I like your thinking.

> Ps. I just pushed a small fix (forgot to check for nulls).
Yeah, I fixed that as well but reverted it because of your change ;)

I've also just [pushed a change to demo-ng](https://github.com/larssn/nativescript-plugin-firebase/blob/8eb96b6760aa1327b520d80d629d70a01710e1d7/demo-ng/app/tabs/firestore/firestore.component.ts#L29-L37) to be able to test the auth change listener more easily. And it works ferfectly fine as far as I can tell 💪

larssn:
> I was going to check that as well. If it doesn't matter for the end-product then I don't mind at all. Perhaps it could be a "dev dependency" even(?)

That was my initial hope, and it compiles just fine with `firebase` as a dev dep, but whatever consumes the plugin will get errors about missing types, unfortunately.

There might be a compromise in that I can probably import the types from the `firebase-webapi.d.ts` file. The downside is that we then have to maintain the types manually to keep them aligned with the official releases, opposed to getting compile time errors when we build the plugin (if they change something, which is preferred IMO).

I'll check out the bundle size, with and without the dependency.

larssn:
So I have a few concerns with the typings, and I think we need to discuss what's the best solution.

First, consumers of the plugin cannot import types from a `d.ts` file. Typescript just refuses to compile when you try to do so, because it's looking for actual code that mirrors the `d.ts` file; so that rules out my suggestion about importing from `firebase-webapi.d.ts`.

So as I see it, that leaves us with two choices:
1. We keep the `firebase` dependency.
2. We create our own interfaces/namespaces/classes so the structure exactly mirrors the `firebase` module.

I'm hovering more towards 1, because the other solution is a massive amount of work to create and maintain. However 1 is more quirky in the fact that people now download a plugin called `nativescript-plugin-firebase`, which exports types from a different plugin called `firebase`, which is not very intuitive consumers of the plugin. We could probably get away with this by the reasoning of "people wanting to use the web API must be used to using the `firebase` dependency anyway". So we'd need to create some documentation that states: "Hey if you're importing from `nativescript-plugin-firebase/app`, you're gonna get the types from `firebase`.

So after writing this, I've convinced myself that 1 is the best approach. :-)

Ps. The bundle size is not affected, because the plugin basically uses 0 lines of javascript from it.

larssn:
Should I include some docs that state that the web API is moving towards the `firebase` dependency? Or are you leaning more towards creating our own interfaces?

larssn:
Bump @EddyVerbruggen

larssn:
I'll figure something out to make the PR more digestible.

Issue #518, opened 2017-10-12
Background Messages Android


bpeterman:
@EddyVerbruggen were you ever able to get background messages to come through? Digging through the code I don't see why it wouldn't work.

To be more specific the issue I'm seeing is when sending a data message (as described here https://firebase.google.com/docs/cloud-messaging/concept-options#data_messages) it doesn't seem to be invoking the onMessageReceived() method when the app is in the background.

Any details you can recall as you were implementing this would be helpful as I'm trying to implement a fix for this.

smarza:
I'm having the same issue.

The onMessageReceivedCallback() is only get called if the APP is running in focus.

jeremypele:
Same here but only tested on ios

richarddavenport:
I have this working currently, in both the init callback and the `firebase.addOnMessageReceivedCallback()`. Here is my manifest
```
<?xml version="1.0" encoding="utf-8"?>
<manifest
    xmlns:android="http://schemas.android.com/apk/res/android" package="__PACKAGE__" android:versionCode="1" android:versionName="1.0">
    <supports-screens android:smallScreens="true" android:normalScreens="true" android:largeScreens="true" android:xlargeScreens="true"/>
    <uses-sdk android:minSdkVersion="17" android:targetSdkVersion="__APILEVEL__"/>
    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
    <uses-permission android:name="android.permission.INTERNET"/>
    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
    <uses-permission android:name="android.permission.SEND_SMS"/>
    <uses-permission android:name="android.permission.RECEIVE_SMS"/>
    <uses-permission android:name="android.permission.READ_SMS"/>
    <uses-permission android:name="android.permission.READ_PHONE_STATE"/>
    <application android:name="com.tns.NativeScriptApplication" android:allowBackup="true" android:icon="@drawable/icon" android:label="@string/app_name" android:theme="@style/AppTheme">
        <activity android:name="com.tns.NativeScriptActivity" android:label="@string/title_activity_kimera" android:configChanges="keyboardHidden|orientation|screenSize" android:theme="@style/LaunchScreenTheme">
            <meta-data android:name="SET_THEME_ON_LAUNCH" android:resource="@style/AppTheme" />
[...]

PR #1056, opened 2018-12-03
Fix Android Background Notification

cerireyhan:
Background notification should be processed in the resumeEvent for android, thats why it wasn't triggering onMessageReceived callback.
https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/518


EddyVerbruggen:
Although I don't have this problem I will merge and test it. Thanks!

cerireyhan:
Thanks! I didn't had this problem until couple of months ago then I always had to get the background notification data from resume-event for android but hopefully this fixed it :)

EddyVerbruggen:
Great, thanks agina. I've now tested and changed a small thing: when a notifications is received in the background and you launch the app from the notification center on Android, the notification callback fires. Then push the app to the background, and bring it forward again, the notification callback fires again. I'm now actively clearing the notification to avoid this.

PR #1064, opened 2018-12-07
Make signInWithEmailAndPassword resolve and reject consistent with web api

BrianChapman:
Passes tests, should be backwards compatible. 

The error message from firebase.login contains additional formatting that isn't consistent with the web api. For example firebase.login formats the error as: 
"Logging in the user failed. com.google.firebase.auth.FirebaseAuthInvalidCredentialsException: The password is invalid or the user does not have a password."
but the web api's error message would just be "The password is invalid or the user does not have a password." Changing the error message could be considered breaking so I didn't make that change.

See https://firebase.google.com/docs/reference/js/firebase.auth.Auth#signInWithEmailAndPassword for web api docs.

### DIFF
--- a/src/app/auth/index.ts
+++ b/src/app/auth/index.ts
@@ -20,9 +20,20 @@ export module auth {
               resolve();
             })
             .catch(err => {
+              let code = 'auth/exception';
+              let message = err.toString();
+              if (message.includes('com.google.firebase.auth.FirebaseAuthInvalidCredentialsException')) {
+                code = 'auth/wrong-password';
+              } else if (message.includes('com.google.firebase.auth.FirebaseAuthInvalidUserException')) {
+                code = 'auth/user-not-found';
+              } else if (message.includes('com.google.firebase.auth.FirebaseAuthInvalidUserException')) {
+                code = 'auth/invalid-email'
+              }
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Hey @BrianChapman! Thanks for this. I'm guessing the `com.google.firebase..` exception is Android-specific? Or did you test this on iOS as well?

EddyVerbruggen:
Btw, looks like the `catch` clause was intended to be added to the signIn method, but you've added it to signOut..

BrianChapman:
Yes, not my finest hour. I've fixed the code that was misplaced in signout.

The IOS implementation simply returns the localized description for the error. So there isn't anything substantial here to use in order to identify the code.

I think the issue with error codes would require a deeper change that I've proposed here. One option is for both src/firebase.android.ts and src/firebase.ios.ts implementations to return a subclass of Error when rejecting the Promise so that the web api layer can map the Error class types to codes and extract the message. However that would break the firebase.login api in that the error is no longer a string and the string representation of the error pre-fixes "Error: " to the message.

EddyVerbruggen:
I see what you mean. Let's get this merged, thanks for your contribution! 🍻

PR #1070, opened 2018-12-09
Dynamic links can be enabled without fauth

sud80:
IOS: Avoid crashing when dynamic links is enabled without FIRAuth


EddyVerbruggen:
Thank you sir!

PR #1071, opened 2018-12-10
Register FirebasePluginLifecycleCallbacks on app-launch 

cerireyhan:
(Foreground: false) for foreground notifications until app is launched from a background notification and app-resume event was executed without this fix.


EddyVerbruggen:
Thanks (again)! Let's get this released asap.

cerireyhan:
No problem :) wish I noticed this earlier tho.

EddyVerbruggen:
Happens to all of us. I'm just grateful you did a follow-up to your PR. ❤️

PR #1098, opened 2019-01-08
note about fingerprint importance

bradmartin:
@EddyVerbruggen feel free to change the wording or positioning, but this needs to be relayed for importance. Have ran into this myself and helped others *cough* (@NathanWalker ) in past saying "it works debug but release doesn't" and the release fingerprint wasn't on firebase.

Just something to stick out on importance 💯


EddyVerbruggen:
Can’t stress that enough. Thanks!

PR #1102, opened 2019-01-09
Add note on App Signing

bradmartin:
Fought the hard battle tonight.


EddyVerbruggen:
Oh wow, didn't release a Firebase app with App Signing enabled yet, this is good to know!

PR #1105, opened 2019-01-10
OnDisconnect - SetWithPriority

KkevinLi:
This PR fixes commit: ed83cb4032e4273e020780ef4d4c2af8921b0a05.

I forgot to pass in the priority parameter  for the setWithPriority function...

### DIFF
--- a/src/firebase.android.ts
+++ b/src/firebase.android.ts
@@ -1726,11 +1726,12 @@ class OnDisconnect implements OnDisconnectBase {
               onFailure: exception => reject(exception.getMessage())
             }));
       } catch (ex) {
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Thanks!

Added you as a collaborator now, so if you have similar small issues please don't hesitate to push them directly.

KkevinLi:
Sounds good! What is the workflow for bigger issues/feature implementation?

EddyVerbruggen:
I guess for features it would be better to do it as a PR because I'd like to test those and check the docs, but for small(er) issues I see no problem in pushing them to master. I'll always review anyway.

PR #1111, opened 2019-01-16
Rework webApi queries

KkevinLi:
This PR refactors the webapi query code which allows us to chain filters + pass in eventTypes.

Before we were only able to do `firebaseWebApi.database().ref(path).orderByKey().on("value", onValueEvent")` but now we can do `firebaseWebApi.database().ref(path).orderByKey().equalTo("value).on("child_added", onValueEvent)`

The api calls are backwards compatible, but now calling `on()` will not fire your callback for every event change at that path. Before `on()` listens to value and child changes whereas now it listens only for the given eventType.


KkevinLi:
I've updated the code so that it more strictly follows the Firebase Web API (To test you can copy paste examples from https://firebase.google.com/docs/database/admin/retrieve-data)

Breaking Change:
- To handle errors you MUST pass in an onErrorEvent callback. (`onValueEvent` will not have a `result.error` to check for any more. A `DataSnapshot` is returned which follows the docs)
- To listen for `child_added, child_removed, child_changed, child_moved` you must pass in that event type (See updated docs). Before we listen to all at once which led to multiple duplicate calls.

Branch also includes:
- Remove short imports as they are deprecated in Nativescript 5.2.0 +
- Fix some types where values can be undefined
- Update to database transaction document
- Removed DataSnapshot.ref() as it's did not return a web` DataSnapshots.ref()`. (Still a TODO)
- Add firebase.database.servervalue.timestamp

KkevinLi:
Any update on this PR? Is there something I need to change before this gets pulled in?

EddyVerbruggen:
Oh man.. ehm.. probably not, but I'm struggling to find time to review it.

Issue #1129, opened 2019-01-31
Separation of Environments doesn't appear to work for iOS


RohanTalip:
I have both of these files:
- app/App_Resources/iOS/GoogleService-Info.plist.dev
- app/App_Resources/iOS/GoogleService-Info.plist.prod

I ran `tns platform remove ios` and then `tns platform add ios` to ensure that the correct files are copied.

Running `tns run ios --bundle` builds the iOS app and starts it in the simulator but the app crashes at this line:
```
const firebase = require("nativescript-plugin-firebase");
```

When I rename app/App_Resources/iOS/GoogleService-Info.plist.dev to app/App_Resources/iOS/GoogleService-Info.plist, delete app/App_Resources/iOS/GoogleService-Info.plist.prod and run `tns platform remove ios`, then `tns platform add ios` and then `tns run ios --bundle`, the iOS app is built and doesn't crash at the require statement.

RohanTalip:
I should point out that I don't have Android enabled or configured for this project, which is why something like pull request #1140 is necessary.

edusperoni:
I'm not sure that PR would work correctly. I was going to do a PR some time ago but things came up and I never got the time. The current behavior copies the google play services only on env change on iOS (so building debug and then release would copy it correctly). I wrote a possible fix but didn't get the time to properly test it: https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/968#issuecomment-432011225


Your PR copy hook could potentially make it rebuild constantly on iOS, or maybe leave undesirable assets behind. I haven't tested it, though.

EddyVerbruggen:
@RohanTalip Thanks, I've merged that PR, and also thanks to @edusperoni, I've added your proposed patch. Let's see with 8.0.0 how this pans out.

rynop:
I have this same problem with version `8.1.1`.  Using NS `5.3.1`. If I have `.dev` extension it errors (below).  If I have a single `.plist` extension file everything works.

```
$ tns run ios
...
nativescript-plugin-firebase: /Users/ryan/projects/firebase-ns-playground/platforms/ios/.pluginfirebaseinfo not found, forcing prepare!
nativescript-plugin-firebase: running release build or change in environment detected, forcing prepare!
nativescript-plugin-firebase: /Users/ryan/projects/firebase-ns-playground/App_Resources/iOS/GoogleService-Info.plist does not exist. Please follow the installation instructions from the documentation

[...]

PR #1140, opened 2019-02-07
Call writeGoogleServiceCopyHook() for both iOS and Android.

RohanTalip:
This should fix issue #1129


EddyVerbruggen:
Good catch, thanks!

Issue #1134, opened 2019-02-05
Custom logs for crashlytics


kanclalg:
Hi, I wonder if custom logs for crashlytics (https://firebase.google.com/docs/crashlytics/customize-crash-reports#ios) supposed to work in the latest plugin version?
I get it to work for android and seems like Crashlytics.log (https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/src/crashlytics/crashlytics.d.ts#L7) works as expected for android but on iOS I don't get any log information on Crashlytics exception "Logs" tab. Looks like the wrong function is called in the code. In the documentation they write that you should call CLS_LOG though I'm not sure how to access this one. Thank you for any help.

kanclalg:
It's documentated in this link under "Add custom logs" https://firebase.google.com/docs/crashlytics/customize-crash-reports#ios

EddyVerbruggen:
Thanks. I doubt that macro is available on NativeScript. Can you try `console.log("CLS_LOG: " + CLS_LOG)`? In case you get a TypeScript error, add `declare const CLS_LOG: any`.

kanclalg:
Yeah, I've already tried it before, it throws:
`ERROR ReferenceError: Can't find variable: CLS_LOG`.
Do you have any ideas how to wrap that macro so it would be accessible to nativescript?

EddyVerbruggen:
I don't think you can, but I haven't extensively googled it. Perhaps we can use these alternatives: https://github.com/irccloud/ios/blob/master/Crashlytics.framework/Headers/CLSLogging.h

kanclalg:
What do you have in mind exactly? None of these seems to be accessible.

EddyVerbruggen:
OK, then it's not possible I guess. But the generated typings seem to suggest `logEvent` is a thing, so I'm not sure why your events aren't logged ☹️

https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/14d83d8cb881f197ec5a4009f0126286fed69041/src/platforms/ios/typings/objc!Crashlytics.d.ts#L154

kanclalg:
`logEvent` is also missing from their official documentation here: https://firebase.google.com/docs/reference/ios/crashlytics/api/reference/Classes/Crashlytics#-recorderror
and looks like it's deprecated: https://github.com/irccloud/ios/blob/master/Crashlytics.framework/Headers/Crashlytics.h#L202
[...]

PR #1142, opened 2019-02-12
Custom TNSCrashlyticsLogger framework for logging custom logs to crashlytics

kanclalg:
Solution for #1134

@EddyVerbruggen I wrote to firebase support regarding if they have any other wrapper function to call `CLS_LOG` and their answer was that `CLS_LOG` is the only way to have custom logs in Crashlytics. As `CLS_LOG` macro is not available from nativescript, I wrote custom framework which just wraps `CLS_LOG` into simple Log function.


kanclalg:
@EddyVerbruggen, can you take a look at it?

EddyVerbruggen:
This looks good, let me test it a little..

EddyVerbruggen:
Excellent work!

PR #1154, opened 2019-02-21
Allow retrieving metadata with queries

acrollet:
This PR puts the basic pieces in place to be able to determine if a result is from cache or not. I found this capability necessary for a project where multiple users might be trying to create a document with a given property that must be unique, and device blips were causing duplicate documents.


EddyVerbruggen:
Hey @acrollet, thanks for this!

While looking at the Firebase Web API I saw some additional features I wanted to immediately add:

1. in addition to `fromCache` you'll now also get the `hasPendingWrites` metadata property.
2. not only queries, but also document snapshots now support passing in the `includeMetadataChanges` flag.

👍❤️

acrollet:
Awesome! Thanks again for your hard work on this plugin 😄

PR #1176, opened 2019-03-14
Firestore onSnapshot update to support onError Callbacks.

KkevinLi:
This PR updates Firestore: document + query onSnapshot so that it supports onError callbacks.

The changes are backwards compatible and allows users to optionally pass in onError() if they wish to handle errors that may occur.

Also fixes a crash that will occur if the first parameter passed to `onSnapshot` was type `firestore.SnapshotListenOptions` and no other parameters were passed in.


EddyVerbruggen:
Supercool, thanks!

PR #1194, opened 2019-03-26
Add methods to set analytics session parameters

mmeven:
In the Analytics module, add a method to set the delay before an analytics session starts :

```setMinimumSessionDuration(milliseconds: number): void;```

Source: [Android](https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.html#setMinimumSessionDuration(long)) & [IOS](https://firebase.google.com/docs/reference/swift/firebasecore/api/reference/Classes/AnalyticsConfiguration#/c:objc(cs)FIRAnalyticsConfiguration(im)setMinimumSessionInterval:)

And another method to set the delay before an analytics session is finished : 

```setSessionTimeoutDuration(milliseconds: number): void;```

Source: [Android](https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.html#setSessionTimeoutDuration(long)) & [IOS](https://firebase.google.com/docs/reference/swift/firebasecore/api/reference/Classes/AnalyticsConfiguration#/c:objc(cs)FIRAnalyticsConfiguration(im)setSessionTimeoutInterval:)

### DIFF
--- a/src/analytics/analytics.android.ts
+++ b/src/analytics/analytics.android.ts
@@ -171,3 +171,15 @@ export function setAnalyticsCollectionEnabled(enabled: boolean): void {
       appModule.android.currentContext || com.tns.NativeScriptApplication.getInstance()
   ).setAnalyticsCollectionEnabled(enabled);
 }
+
+export function setMinimumSessionDuration(milliseconds: number): void {
+  com.google.firebase.analytics.FirebaseAnalytics.getInstance(
+      appModule.android.currentContext || com.tns.NativeScriptApplication.getInstance()
+  ).setMinimumSessionDuration(milliseconds);
+}
+
+export function setSessionTimeoutDuration(milliseconds: number): void {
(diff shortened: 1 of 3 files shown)


mmeven:
I've removed the ```setMinimumSessionDuration``` as it becomes deprecated in the latest versions of the Analytics SDK (in both Android and iOS). The other method is still relevant if I understand correctly.

EddyVerbruggen:
Cool, thanks! I'll add a few words to the docs.

EddyVerbruggen:
Just noticed according to https://github.com/firebase/firebase-ios-sdk/pull/2516 `setSessionTimeoutInterval` is moving to a new spot in the SDK, so when bumping to iOS SDK 5.19.0 we need to refactor this.

EddyVerbruggen:
Also, I wonder if iOS also uses milliseconds (like Android does), as the doc you linked to mentions seconds.

PR #1201, opened 2019-03-29
Using Non-Firebase Push Messaging with other Firebase Services

alfonsocj:
Adds the ability to have Non Firebase Push Messaging along with other Firebase services.

if `external_messaging` in `firebase.nativescript.json` is true, then it won't add the Firebase/Messaging iOS SDK. 

We keep `"messaging": true` in `firebase.nativescript.json` to enable Android push messaging.

### DIFF
--- a/src/scripts/postinstall.js
+++ b/src/scripts/postinstall.js
@@ -4574,7 +4574,7 @@ post_install do |installer|
 end`) + `
 
 # Firebase Cloud Messaging (FCM)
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Thanks! I've added a note to the [docs](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/docs/NON_FIREBASE_MESSAGING.md) on how to use this.

PR #1205, opened 2019-04-04
Web auth signin

KkevinLi:
This PR adds auth sign in with a custom token and via email link. It also cleaned up missing semicolons.


EddyVerbruggen:
Code looks good, I'll test this soon! Thx!

PR #1217, opened 2019-04-08
[auth] - Update User to match web api interface

KkevinLi:
This PR renames some properties of User to better follow the Web API User Interface. Since this can break users I decided to put it on branch.

https://firebase.google.com/docs/reference/js/firebase.User


EddyVerbruggen:
I don't mind too much. Adding a note to the changelog and docs. Cheers mate!

PR #1218, opened 2019-04-09
fix: onAdClosed it's never fired

adriexnet:
This PR fixes the callback onAdClosed, that it's not working with `preloadInterstitial` on ios because the delegate is cleaned as soon the ad it's loaded so the onAdCloseCallback will never be fired


EddyVerbruggen:
Makes sense, thanks! 👍

PR #1221, opened 2019-04-10
fix: preloadInterstitial promise doesn't resolve for a second ad

adriexnet:
This PR fixes the preloadInterstitial promise, that it's not working on android because of an issue related to JavaScript closures that not allow resolve multiple promises, it's the same issue that https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/src/admob/admob.android.ts#L30


EddyVerbruggen:
THANKS!

Issue #1222, opened 2019-04-10
application.currentContext is deprecated


000panther:
When using the plugin, I receive a lot of status messages in the log:

application.currentContext is deprecated; use startActivity, foregroundActivity, or context instead

Since deprecated stuff tends to be removed some day, it would be cool to replace the reference as described in the log message.

Github sees 2 references in the code: 

**https://github.com/EddyVerbruggen/nativescript-plugin-firebase/search?q=currentContext&unscoped_q=currentContext**

PR #1223, opened 2019-04-10
This would fix the issue #1222 with deprecated currentContext

000panther:
Not sure if it has compability issues with lower nativescript versions though, if yes, I could add some handling code. It works on my system with nativescript 

```
✔ Component nativescript has 5.3.1 version and is up to date.
✔ Component tns-core-modules has 5.3.1 version and is up to date.
✔ Component tns-android has 5.3.1 version and is up to date.
```

https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/1222


EddyVerbruggen:
Thanks! Actually use `appModule.android.context` already in other parts of the codebase, so I don't think this will do harm. 👍

PR #1229, opened 2019-04-17
fix android token generation workflow

### LINKED ISSUE #1045: Messaging: onPushTokenReceivedCallback called immediately after unregisterForPushNotifications
On Android, right after I successfully have called messaging.unregisterForPushNotifications(), the callback onPushTokenReceivedCallback gets called returning a new token. So the phone gets unregistered successfully but then receives a new token.

I have tested this on a Google Pixel 2, Samsung Galaxy S8 and S4 (real devices).

I have tried creating a new project and added this:

`require("nativescript-plugin-firebase");`

in app.js

and this in main-view-model.js:

`viewModel.register = function () {
        messaging.registerForPushNotifications({
            onPushTokenReceivedCallback: (token) => {
              console.log("Firebase plugin received a push token: " + token);
            },
          
            onMessageReceivedCallback: (message) => {
              console.log("Push message received: " + message.title);
            },
          
            // Whether you want this plugin to automatically display the notifications or just notify the callback. Currently used on iOS only. Default true.
            showNotifications: false,
          
            // Whether you want this plugin to always handle the notifications when the app is in foreground. Currently used on iOS only. Default false.
            showNotificationsWhenInForeground: false
          }).then(() => console.log("Registered for push"));
    }

    viewModel.unRegister = function () {
        messaging.unregisterForPushNotifications().then(() => {
              console.log("Unregistered!!!");
        });
    }`

How to avoid this? Or what am I doing wrong?

I'm using {N} 5 and 7.4.2 of the firebase-plugin.

### LINKED ISSUE #1228: [Push Only] Workflow difference between iOS and Android
In my app I have a switch and allows the user to enabled/disable push notifications. By default when the app opens for the first time, it enables push:
```ts
    messaging.registerForPushNotifications({
        onMessageReceivedCallback: (message) => {
            console.log(message);
        },
        onPushTokenReceivedCallback: (token) => {
            console.log(token);
        }
    })
``` 
If the user taps on the switch I disable notifications by calling `messaging/unregisterForPushNotifications()`. Then the user can tap again on the switch to re-enable notifications and this calls the same method that is called on initial app start. 

For iOS this works ok and the `onPushTokenReceivedCallback` is called everytime I call register. For Android though things do not work like that. The callback is triggered only on initial app load. After that, no matter how many times I call register the callback is never triggered. 

Seems related with the other problem described in #1045

PeterStaev:
Fixes #1045 
Fixes #1228

### DIFF
--- a/src/platforms/android/libraryproject/firebase/src/main/java/org/nativescript/plugins/firebase/FirebasePlugin.java
+++ b/src/platforms/android/libraryproject/firebase/src/main/java/org/nativescript/plugins/firebase/FirebasePlugin.java
@@ -11,6 +11,7 @@ public class FirebasePlugin {
 
   private static String cachedToken;
   private static String cachedNotification;
+  private static boolean preventInitialRegisterTokenCallback = false;
 
   private static FirebasePluginListener onPushTokenReceivedCallback;
   private static FirebasePluginListener onNotificationReceivedCallback;
@@ -19,7 +20,13 @@ public class FirebasePlugin {
     new Thread() {
       public void run() {
         try {
-          FirebaseInstanceId.getInstance().getToken(senderId, "FCM");
+          String token = FirebaseInstanceId.getInstance().getToken(senderId, "FCM");
+
+          if (!preventInitialRegisterTokenCallback) {
+            executeOnPushTokenReceivedCallback(token);
+          }
+
+          preventInitialRegisterTokenCallback = false;
         } catch (IOException e) {
           Log.e(TAG, "Error getting a token from FCM: " + e.getMessage(), e);
         }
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
That's a very welcome change - I will test this and release it with 8.3.0. Thank you!

IAMtheIAM:
I look forward to this release

PeterStaev:
Hey @EddyVerbruggen , I see version 8.3 was released but from my tests seems the code to the native android side have not been bundled. Can you please check and see if the build has the `firebase-release.aar` updated? I really need this change so I can publish an update to an app that still uses GCM and will stop working after May 29th...

EddyVerbruggen:
Oh.. that would be disappointing. I'll take a look later today and publish a new version if necessary!

EddyVerbruggen:
Your assessment was spot on. I hope you fare better with (the now published) 8.3.1.

PeterStaev:
Thanks Eddy, works like a charm now!

PR #1250, opened 2019-05-02
use fully qualified VersionNumber

vtrifonov:
As we [plan to release using gradle 5.1.1](https://github.com/NativeScript/android-runtime/issues/1360) and VersionNumber is not resolved there I'm adding the namespace in front.


EddyVerbruggen:
Thanks!

PR #1252, opened 2019-05-04
Documented crashlytics.log() function.

tralves:
This PR documents the `crashlytics.log()` function. It is very useful and it was a very nice surprise to see it was in fact implemented (and works)!

I didn't document the optional parameters (tag and priority) because it is confusing and works differently in different platforms. It seems to me that these docs choose to simplify and focus on the "happy path".


EddyVerbruggen:
Thanks! I think that was missing from #1134 👍

PR #1257, opened 2019-05-09
fix: replace deprecated CLI's logger function

rosen-vladimirov:
In CLI 5.4.0 `$logger.out` will be deprecated, so use `$logger.info` instead.

More information is available in [this issue](https://github.com/NativeScript/nativescript-cli/issues/4602)


EddyVerbruggen:
Great changes, and thanks for this PR!

PR #1258, opened 2019-05-13
fix: getting merge assets output folder

vtrifonov:
In the new android gradle plugin 3.4.0 the location of the merge assets is moved again in a different folder than the one from 3.3.2, so we did some changes not to rely on hardcoded paths, but to get the path from the task's output folder.

### DIFF
--- a/src/scripts/postinstall.js
+++ b/src/scripts/postinstall.js
@@ -5270,29 +5270,27 @@ module.exports = function($logger, $projectData) {
             appBuildGradleContent = appBuildGradleContent.replace("ensureMetadataOutDir.finalizedBy(buildMetadata)", "ensureMetadataOutDir.finalizedBy(buildMetadata)\\n\\t\\tbuildMetadata.finalizedBy(copyMetadata)");
             appBuildGradleContent += \`
 task copyMetadata {
-  doLast {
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Hey @vtrifonov, thanks for helping out, much appreciated!

As discussed on Slack, I get this error: `Unable to apply changes on device: emulator-5554. Error is: project is not defined.`

Note: I've applied your changes to the source file of the script you changed ([this file](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/publish/scripts/installer.js)). Sorry for the confusion. I'd rather not publish the generated file (you changed) to avoid confusion, but for "reasons" I need this file to be in here.

Anyway, if you want to run the demo app and reproduce the error I'm seeing after applying this pactch, do the follwing:

```bash
git clone https://github.com/EddyVerbruggen/nativescript-plugin-firebase
cd nativescript-plugin-firebase
cd src
npm i # this will trigger a bunch of prompts which you can answer 'n' to (the result is ignored)
npm run demo.android
```

That should give you the error I am seeing, unless your local environment has some other setup (gradle or otherwise) that doesn't trigger the issue.

PR #1260, opened 2019-05-14
fix: use string concatenation instead of format

vtrifonov:
The problem with the `project is not defined` comes because of using ${} which has to be escaped twice, because we are doing string replace twice, but to avoid this I've changed it with using string concatenation instead of string format.

### DIFF
--- a/publish/scripts/installer.js
+++ b/publish/scripts/installer.js
@@ -1115,10 +1115,10 @@ task copyMetadata {
     copy {
       if (!project.mergedAssetsOutputPath) {
         // mergedAssetsOutputPath not found fallback to the default value for android gradle plugin 3.4.0
-        project.ext.mergedAssetsOutputPath = "$projectDir/build/intermediates/assets/\${project.selectedBuildType}/out"
+        project.ext.mergedAssetsOutputPath = "$projectDir/build/intermediates/assets/" + project.selectedBuildType + "/out"
       }
-        from "$projectDir/src/main/assets/metadata"
-      into "\${project.mergedAssetsOutputPath}/metadata"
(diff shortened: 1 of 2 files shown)


EddyVerbruggen:
Smart move, thanks!

EddyVerbruggen:
Just noticed the build fails on Android when using tns-android 5.1.0. With 5.2.0 all is fine though.

The problem is on this line:

https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/db107236ca4a6733012e58fa2713ce978ebb2806/publish/scripts/installer.js#L1103

Error: `> Could not find method getMergeAssetsProvider() for arguments [] on object of type com.android.build.gradle.internal.api.ApplicationVariantImpl.`

I'll try to make it more backward compatible.

vtrifonov:
Can you try with the following code:
```groovy
android.applicationVariants.all { variant ->
    if (variant.buildType.name == project.selectedBuildType) {
        def task
        if (variant.metaClass.respondsTo(variant, "getMergeAssetsProvider")) {
            def provider = variant.getMergeAssetsProvider()
            task = provider.get();
        } else {
            // fallback for older android gradle plugin versions
            task = variant.getMergeAssets()
        }
        for (File file : task.getOutputs().getFiles()) {
            if (!file.getPath().contains("/incremental/")) {
                project.ext.mergedAssetsOutputPath = file.getPath()
            }
        }
    }
}
```

EddyVerbruggen:
I just learned about that `respondsTo` trick as well and was creating something very similar to your snippet. Yours seems to work perfectly fine, so using that. Cheers!

PR #1261, opened 2019-05-14
fix: make the plugin androidx compatible

tgpetrov:
As this plugin uses the old Android Support Libraries, it will not be compatible with the upcoming NativeScript 6.0 release which will use [AndroidX by default](https://www.nativescript.org/blog/support-for-androidx-in-nativescript). The changes from this PR, make the plugin compatible with the AndroidX library (NS 6.0) and with the old libraries (NS before 6.0).


EddyVerbruggen:
Looks great! This will be part of 8.3.2 which I plan to release later today. 👍

PR #1265, opened 2019-05-21
feat(google): include web friendly oauth2 token

NathanWalker:
I was recently working on a shared mobile/web project where the backend needed to use the web oauth2 token but it wasn't provided. This makes it available and easy to access 👍


EddyVerbruggen:
That's great, thanks Nathan!

PR #1272, opened 2019-05-24
Use notification shared delegate on iOS

edusperoni:
This is a mirror implementation from `nativescript-local-notifications`.

Additional testing needed (especially with non firebase push).

Fixes https://github.com/EddyVerbruggen/nativescript-local-notifications/issues/107


saeb-panahifar:
Hi , Thank you for resolving this issue. I use it in my project. It works perfectly.

edusperoni:
Thanks @saeb-panahifar for testing it!

@EddyVerbruggen I believe this and https://github.com/EddyVerbruggen/nativescript-local-notifications/pull/155 are ready for merging.

PeterStaev:
@edusperoni , I haven't specifically tested your code, but don't think it will solve all the problems. Looking at it you have changed just the `messaging.ios.ts` file. But in my case if the app is fully closed (not in background) having the initial setup of the plugin in `app.ts` - `require("nativescript-plugin-firebase")`. So seems there is something in that file that messes up local notifications when the app is closed. I ended up adding the following code (instead of the recommended setup)
```ts
import { messaging } from "nativescript-plugin-firebase/messaging";

if (global.isIOS) {
    if (app.ios.delegate === undefined) {

        @ObjCClass(UIApplicationDelegate)
        class UIApplicationDelegateImpl extends UIResponder implements UIApplicationDelegate {
        }

        app.ios.delegate = UIApplicationDelegateImpl;
    }
    messaging.addBackgroundRemoteNotificationHandler(app.ios.delegate);
}
if (global.isAndroid) {
    // tslint:disable-next-line: no-var-requires
    require("nativescript-plugin-firebase");
}
```

Have you validated your fix with such a behavior - having the app closed (not in background), trigger a local notification and then tapping on the notification will trigger the local notification callback.

edusperoni:
@saeb-panahifar tested the notifications when the app was closed (swiped off) and it worked. You're welcome to test it too, and if needed we can add some extra code to handle background notifications.

The idea behind this approach is that any developer that wants to handle notifications should use my lib to ensure no conflicts. Ideally this should be handled by helpers in the core.

This PR removes the part where the plugin registers itself as the delegate and instead observes a shared delegate that registers itself on construct, so any extra registration should be done in `nativescript-shared-notification-delegate`.

src/messaging/messaging.ios.ts
```diff
@@ -551,7 +547,13 @@ class UNUserNotificationCenterDelegateImpl extends NSObject implements UNUserNot
...
 
-  public userNotificationCenterDidReceiveNotificationResponseWithCompletionHandler(center: UNUserNotificationCenter, response: UNNotificationResponse, completionHandler: () => void): void {
+  public userNotificationCenterDidReceiveNotificationResponseWithCompletionHandler(center: UNUserNotificationCenter, response: UNNotificationResponse, completionHandler: () => void, next: () => void): void {
+    const userInfo = response.notification.request.content.userInfo;
+    const userInfoJSON = firebaseUtils.toJsObject(userInfo);
+    if (!userInfoJSON["gcm.message_id"]) { // not a firebase message!
+      next();
+      return;
+    }
     // let's ignore "dismiss" actions
```
PeterStaev: This will probably not work for non-firebase messages. I'm using azure notification hubs, and it doesn't have a `gcm.message_id` property. There is a `data.aps` property that might be good to check here as well.

PeterStaev:
I guess Eddy's consideration for not merging this, is that it adds a dependency to another plugin. 

I totally agree that the better approach would be to have this built-in the core modules, but not sure how willing will be the NS team to merge such a thing to core...

src/messaging/messaging.ios.ts
PeterStaev: Actually here is a better way to check if the push is from remote or not (since it doesn't rely on any JSON properties in the payload):
https://stackoverflow.com/a/48717664/5272836

src/messaging/messaging.ios.ts
```diff
@@ -516,27 +517,22 @@ function _addObserver(eventName, callback) {
...
   }
 
-  public userNotificationCenterWillPresentNotificationWithCompletionHandler(center: UNUserNotificationCenter, notification: UNNotification, completionHandler: (p1: UNNotificationPresentationOptions) => void): void {
+  public userNotificationCenterWillPresentNotificationWithCompletionHandler(center: UNUserNotificationCenter, notification: UNNotification, completionHandler: (p1: UNNotificationPresentationOptions) => void, next: () => void): void {
     const userInfo = notification.request.content.userInfo;
     const userInfoJSON = firebaseUtils.toJsObject(userInfo);
+    if (!userInfoJSON["gcm.message_id"]) { // not a firebase message!
+      next();
+      return;
+    }
```
PeterStaev: This check should also be changed similar to the one below so it can work with non-FCM pushes.

edusperoni:
@PeterStaev @saeb-panahifar could you try this fix? I'm not on our dev mac at the moment

PeterStaev:
I can't test it right now either but LGTM!

EddyVerbruggen:
Thanks guys for all your efforts making this happen! This has now been published to npm.

PR #1296, opened 2019-06-13
update postinstall.js to support  Xcode 10

milansar:
Xcode 10 require to have input file field set to point to info.plist file
see https://firebase.google.com/docs/crashlytics/get-started

Have not checked whether it still works with Xcode 9 base Builds
On My machine whenever i build project new build phase Configure Crashlytics added in XcodeProject file


EddyVerbruggen:
Thank you!

PR #1299, opened 2019-06-14
Use instance of `FirebaseStorage` instead of ReferenceFromUrl

### LINKED ISSUE #1298: [Android] Unable to use custom buckets
On Android it throws error when other bucket then the default one.
```
firebase.storage.uploadFile({
            bucket: `gs://my-custom-bucket`,
            onProgress: onProgress,
            remoteFullPath: remotePath,
            localFile: data
        });
```

Error: 

> The supplied bucketname does not match the storage bucket of the current instance.

I have double checked that the custom bucket is in the same Firebase project.
I have not tested if the error is also thrown on iOS

neryortez:
Using a reference from the default instance does not allow to use other storage bucket than the default.
Changes made as documented: https://firebase.google.com/docs/storage/android/start#use_multiple_storage_buckets

Fixes #1298

### DIFF
--- a/src/storage/storage.android.ts
+++ b/src/storage/storage.android.ts
@@ -21,7 +21,7 @@ function getStorageRef(reject, arg) {
     return;
   }
 
-  return arg.bucket ? com.google.firebase.storage.FirebaseStorage.getInstance().getReferenceFromUrl(arg.bucket) : firebase.storageBucket;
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Thanks, but can you please update your fork? It's 600+ commits behind and now we have a merge conflict. Cheers!

neryortez:
I don't understand how it happened xD

Closing un favor or #1326

PR #1304, opened 2019-06-18
fix: fix cannot read .toLowerCase() of undefined

Fatme:
NativeScript CLI throws `cannot read .toLowerCase() of undefined` error as there isn't platform property in platformData object. We should use prepareData instead.


EddyVerbruggen:
Thanks!

PR #1314, opened 2019-06-25
fix: parse correctly CLI's version

rosen-vladimirov:
In case the current Node.js version is not officialy supported by NativeScript CLI, i.e. its new and it is not verified or support for it is deprecated, `tns --version`'s ouput contains warning.
Extract the actual version from the output no matter if there's a warning or not.


EddyVerbruggen:
Good call! Thank you both ❤️

PR #1319, opened 2019-06-29
Add missed optional token property to Provider interface

### LINKED ISSUE #1316: It seam that Provider interface need to extend with token property
This issue related to operation with plugin by TypeScript.

In [Facebook login](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/docs/AUTHENTICATION.md#facebook-login) and [Google Sign-In](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/docs/AUTHENTICATION.md#google-sign-in) documentation mentioned that as result we get response with next part:
``` json
"providers": [
         {
             "id": "facebook.com | google.com",
             "token": "<FB token> | <Google Access Token>"
         }
     ]
```
and I got it in `result` which is `User` interface, but `Provider` interface does not have `token` property, so I need to cast it to `any` to access to `token`.

So it seam that `Provider` interface need to by updated next way:
``` ts
export interface Provider {
  id: string;
  token?: string;
}
```

ycherniavskyi:
Resolve #1316.

### DIFF
--- a/src/firebase.d.ts
+++ b/src/firebase.d.ts
@@ -233,6 +233,7 @@ export interface IdTokenResult {
 
 export interface Provider {
   id: string;
+  token?: string;
 }
 
 export interface FirebasePasswordLoginOptions {


EddyVerbruggen:
Cheers!

PR #1326, opened 2019-07-10
Get Bucket from Instance method

neryortez:
As documented here: https://firebase.google.com/docs/storage/android/start#use_multiple_storage_buckets


EddyVerbruggen:
Thank you! 👍

PR #1346, opened 2019-07-29
Add info to specify push notification icon/color

000panther:
Android Firebase push notifications are configured via the AndroidManifest.xml. Added info how to do this.


EddyVerbruggen:
Interesting find, thank you!

PR #1360, opened 2019-08-15
Added support for inverse barcodes

zalesak:


src/mlkit/barcodescanning/index.d.ts
```diff
@@ -7,11 +7,11 @@ export interface MLKitScanBarcodesResultBounds {
   origin: {
     x: number;
     y: number;
-  },
+  };
```
EddyVerbruggen: Ha, good catch!

EddyVerbruggen:
Really impressive stuff!

Do you perhaps have an example of an inverse barcode you tested your implementation with? I'd love to give it a spin.

zalesak:
Hi Eddy, sorry for late response.
Unfortunately, I have just specimen from drugs packages, from customers. I can`t publish it wide. But in principle, solution just inverse picture before sending it to ML KIT so it will work with every inverse barcode. 
We are using it on 2D matrix and QR Codes.

EddyVerbruggen:
OK, thanks for the reply. Then I'm just going to assume it works ;)

zalesak:
You can find inverted qrcode on google or just invert image in bitmap editor and try it :) e.g.

EddyVerbruggen:
@MurilloBrand You're spot on. Let me release 9.1.2 today, with a rebuilt jar. Thanks!

PR #1362, opened 2019-08-16
Update README for a better Vue/DB example

jlooper:
Just updated mixology-mobile for realtime database and custom models, works with NS6


EddyVerbruggen:
That’s super cool! TY :)

Issue #720, opened 2018-05-21
Camera is stretched when using width and height


Burgov:
I have a barcode scanner with width 400 and height 200, and the camera view is stretch to make everything really flat, looks like it's scaling the full screen camera down to the dimensions without keeping aspect ratio.

EddyVerbruggen:
Yeah, I noticed this as well (only on Android actually).

ndcunningham:
+1

shaqaruden:
I have this issue on iOS as well. Can you fix this? Or is this a Firebase MLKit issue?

EddyVerbruggen:
Can you share the UI code?

shaqaruden:
The code in the example

shaqaruden:
Changing the width and height to 100% does not distort the image

EddyVerbruggen:
So that means it's ok now?

shaqaruden:
Yea, but it would be nice to be able to have a camera view that's not the full size

zalesak:
I hope this is fix: 
In file mlkit-cameraview.android.js, line 243 replace with this:
var diff = Math.abs((size.width / size.height) - (pict.width / pict.height));

it compute preview/picture pair with the closest aspect ratio.

DanielDent:
To the extent that this is an Android-specific issue, I think the following line is responsible:

https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/794467a9d72b66247c1343eee3a7665c6b00306c/src/mlkit/mlkit-cameraview.android.ts#L130

The selectSizePair function seems to be looking for the previewSize which most closely matches the aspect ratio of the display area -- however, it assumes that the display area is always 1400x1200. Identifying the camera preview's current height & width, and updating it based on screen orientation or other layout changes does not currently appear implemented.

EddyVerbruggen:
Yeah that TODO is waiting to be solved indeed. Happy to merge a PR if someone would like to take a stab at it :)

MurilloBrand:
I solved this!
I will create a PR soon

EddyVerbruggen:
@MurilloBrand That would be great!

Issue #1364, opened 2019-08-19
Bump native SDK versions


EddyVerbruggen:
It's that time of the month again.

- iOS: [6.6.0](https://firebase.google.com/support/release-notes/ios#version_660_-_august_6_2019)
- Android: [core 17.0.1, etc](https://firebase.google.com/support/release-notes/android#update_-_august_1_2019)

PR #1365, opened 2019-08-19
Fixing Camera Stretch

MurilloBrand:
This should solve the problem described at #720 
Tested on Android 9


EddyVerbruggen:
@MurilloBrand This seems to work really well, thanks a lot!

I currently run into issues on Android because of the SDK bumps in #1364 so I can't release the new version yet. But I aim to do it one of these days.

EddyVerbruggen:
FYI In 9.1.1 I'm updating this logic a little to make sure the entire viewport is covered by the camera view, because IMO it looks better than cropping the image.

Issue #1341, opened 2019-07-23
MLKitTextRecognition rotation bug when device orientation changed 


jacky-tay:
I've been testing the demo app for text recognition, and I have noticed that when the device orientation changed, live camera preview seems to stay at its previous orientation.

![Bug](https://user-images.githubusercontent.com/1891395/61674913-2adfb580-ad4a-11e9-812f-44dbf66d0bf9.gif)

alexander-mai:
This is a bug in the android implementation.
There is no event listener for orientation change, which corrects the camera orientation like on iOS.
The orientation is only set on once here: https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/4c60759c2148fbcbde09f65868de0a828dc8b37c/src/mlkit/mlkit-cameraview.android.ts#L157
Adding an listener like on iOS will fix the problem:
application.on("orientationChanged", () => { this.setRotation(this.camera, parameters, requestedCameraId) });

PR #1373, opened 2019-08-22
FIXED orientation bug on android

alexander-mai:
* https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/1341


EddyVerbruggen:
THANKS!

I noticed with or without this PR applied in landscape text recognition didn't work. So I've changed a few small things and now it seems to work beautifully. Really happy with this, so a big thank you!

PR #1389, opened 2019-09-04
Fixing dynamicLink crash

nmaswood98:
fixing crash if firebase returns a null dynamic link in its completion handler when handling dynamic links


EddyVerbruggen:
That shouldn't hurt. Thanks!

Issue #1392, opened 2019-09-06
what is the correct way to monitor RealTime Database changes using the plugin?


carlosdelfino:
I wrote the code below to monitor the "/ users" path in Realtime Database, so that whenever a new user is added to this path the application is informed, but I only get data when the code is initialized when the listener is added, ie only once returns all the data, I understand this initial operation, but what is preventing the callback call when new data is added or when some data is changed or removed?

```
    async init() {
        this._settings.debug && console.log("PersistenceService.init");
        return await firebase.init({ persist: true })
            .then((result) => {
                if (this._settings.debug)
                    console.log("PersistenceService.init firebase.init.then():", result);
                firebase.keepInSync("/users", true).then(() => {
                    var onValueEvent = (result: firebase.FBData) => {
                        console.log("Event type: " + result.type);
                        console.log("Key: " + result.key);
                        console.log("Value: " + JSON.stringify(result.value));
                    };
                    firebase.addChildEventListener(onValueEvent, "/users").then(
                        (listenerWrapper: firebase.AddEventListenerResult) => {
                            this._settings.debug && console.log("PersistenceService.init firebase.init.then firebase.addChildEventListener");
                            var path = listenerWrapper.path;
                            var listeners = listenerWrapper.listeners; // an Array of listeners added
                            // you can store the wrapper somewhere to later call 'removeEventListeners'
                        }
                    ).catch((error) => console.log("PersistenceService.init firebase.init.then firebase.addChildEventListener.catch", error));
                });

                return result;
            })
            .catch((error) => {
                if (this._settings.debug)
                    console.log("AuthService.new(): firebase.init.catch()", error);
                throw error;
            })
    }
```

The codes below work perfectly:

```
   /** */
    registerStatus(user: IslamicUser): void {
        firebase.setValue("/users/" + user.uid, user);
    }

    /** */
    updateUserPosition(user: IslamicUser, position: GPSInfo) {
        if (this._settings.debug)
            console.log("PersistenceService.updateUserPosition", user, position);
        position.updated = ServerValue.TIMESTAMP;
        firebase.setValue("/users/" + user.uid + "/position", position)
            .then(() => {
                if (this.debug) console.log("PersistenceService.updateUserPosition firebase.setValue");
                return firebase.getValue("/users/" + user.uid + "/position/updated")
                    .then((snapshot) => {
                        if (this._settings.debug)
[...]

PR #1400, opened 2019-09-11
Adding information regarding access rules.

carlosdelfino:
If the listeners are used, they are indicated to be added after
authentication to Firebase, if the rules prevent access to data without
authentication. relative Issue #1392

TODO: Improve handling of rule-related errors by giving more details of
what caused the error if possible.

TODO: Check if there is a way to avoid such an access problem.

### DIFF
--- a/docs/DATABASE.md
+++ b/docs/DATABASE.md
@@ -256,8 +256,11 @@ Removes values in multiple locations
 ```
 
 ### addChildEventListener
+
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
Thx! Because of the todo's you added: should I wait merging?

EddyVerbruggen:
👍

PR #1428, opened 2019-10-10
Update firebase.android.ts

Langers8:
Line 302: Replace '===' with '!=='?
Code causes error if analytics disabled.
Is this an error or should 'com.google.firebase.analytics.FirebaseAnalytics' be defined within the condition before using 'setAnalyticsCollectionEnabled'?


EddyVerbruggen:
Oh jeez this is an error indeed. Good catch, thanks!

Issue #933, opened 2018-09-29
firebase.init error: TypeError: Cannot read property 'getInstance' of undefined


ShaimaaKhamees:
i try to use firebase plugin in my nativescript core(not angular) project
and i noticed that reference.d.ts file not exist in my project to put this line
"<reference path='./node_modules/nativescript-plugin-firebase/firebase.d.ts'/>"
i try to install and reinstall the plugin and remove /add platform android without any changes and i stuck in this step when i put the following code

firebase.init({
        // Optionally pass in properties for database, authentication and cloud messaging,
        // see their respective docs.
    }).then(
        instance => {
            console.log("firebase.init done");
        },
        error => {
            console.log(`firebase.init error: ${error}`);
        }
    );

any suggestions please

tomylee001001:
i have the same problem.
i install my plugin normally and i have "firebase.nativescript.json" file created at my root by installer script.
any help please

agarzas:
Same problem here, I´m also using nativescript angular 6.1 and nativescript plugin firebase 7.1.1...

I tried removing the platform/android, reinstalling the plug-in without success.

felix-idf:
I have the same problem using NativeScript 4.2 without Angular and nativescript-plugin-firebase 7.1.1.
No success after removing platform android, re-installing the plugin and building the app with tns run android --syncAllFiles.

EddyVerbruggen:
Share a repo that easily shows the issue please.

felix-idf:
[PluginFireBaseIssue.zip](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/files/2437072/PluginFireBaseIssue.zip)

Just open the app, tap the button to call the init method and you see the message 

> TypeError: Cannot read property 'getInstance' of undefined

EddyVerbruggen:
Thanks, @felix-idf, that really helps. I don't think I've tested recent versions with this `firebase.nativescript.json` config (only `using_android` and `messaging` are `true`):

```json
{
    "external_push_client_only": false,
    "using_ios": false,
    "using_android": true,
    "firestore": false,
    "realtimedb": false,
    "authentication": false,
    "remote_config": false,
    "messaging": true,
    "crashlytics": false,
    "crash_reporting": false,
    "storage": false,
    "functions": false,
    "facebook_auth": false,
    "google_auth": false,
    "admob": false,
    "invites": false,
    "dynamic_links": false,
    "ml_kit": false
}
[...]

PR #1435, opened 2019-10-18
Make sure FirebaseAnalytics is defined to prevent firebase.init from …

JohanBeumer:
…failing when analytics is not enabled.

Just like the solution for #933 .


EddyVerbruggen:
Thanks! That won't hurt.

PR #1437, opened 2019-10-18
[DOC] - add a couple of notes to messaging

sbatson5:
## What's in this PR?

This just adds a couple of notes to the `messaging` section.  We recently shipped a NativeScript app to the app store and there were a couple of minor details that were a bit confusing.

We could not get push notifications to work without a `onMessageReceivedCallback` -- after my first reading, I thought that was an optional function for doing something within the app when a message is received.  I didn't think that notifications wouldn't work without it.

Also added a note that the entitlements file needs to be updated for production when you are ready to ship to the app store.

### DIFF
--- a/docs/MESSAGING.md
+++ b/docs/MESSAGING.md
@@ -43,6 +43,13 @@ so it's not removed when you remove and re-add the iOS platform. The relevant co
 
 > Note that since plugin version 5.1.8 the name of the file can either be `<YourAppName>.entitlements` or `app.entitlements`. `YourAppName` is the iOS folder name, see the path above.
 
+> Note that when shipping to the iOS App Store or Google Play, the entitlements file needs to be updated to reflect production:
+
+```xml
+	<key>aps-environment</key>
+	<string>production</string>
+```
+
 #### Allow processing when a background push is received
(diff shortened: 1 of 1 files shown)


docs/MESSAGING.md
```diff
@@ -144,6 +153,7 @@ parameters/data when sending the push notification, you need to set the `showNot
 
 ```js
 firebase.init({
+  // ..
```
sbatson5: Just added these because I had the impression this might be all I need to get things working

EddyVerbruggen: I'm not sure what you mean.. can you please explain?

sbatson5: Sorry, what I meant was that when I saw just:
```
firebase.init({
  showNotifications: true,
});
```
Without the `// ...` I thought that maybe that was all I needed to get things working.  The `// ...` implies there are other things that have to be in there. Maybe that's just me but having just this flag won't really do anything 🤷‍♂

EddyVerbruggen: Ah, gotcha. Yep you're right. Makes sense.

EddyVerbruggen:
Thank you!

PR #1438, opened 2019-10-21
getFunctions should have had the region as its argument on iOS

ibnYusrat:
`getFunctions` should have had the region as its argument on iOS, which was not there, and subsequently caused problem accessing cloud functions on iOS if they are in a different region than `us-central`.


EddyVerbruggen:
Oh man, that's a bug right there! Thx!

PR #1463, opened 2019-11-04
Return the GIDSignIn id token when login with Google on iOS

francoisph:
The token returning from the login with Google on iOS  is the same on both platforms : firebase._googleSignInIdToken on android and GIDSignIn.sharedInstance().currentUser.authentication.idToken on iOS.

https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/1455


EddyVerbruggen:
Much better, thank you!

Resolves #1455.

PR #1474, opened 2019-11-08
Avoid app crash when googleSignInResult is null

### LINKED ISSUE #1473: Android googleSignInResult can be null and crash the app
Hello,

Crashlytics report a lot of "Cannot read property 'isSuccess' of null" while android google sign in. (Just after `screen_view firebase_screen_class: SignInHubActivity`)

It seems that `const googleSignInResult = com.google.android.gms.auth.api.Auth.GoogleSignInApi.getSignInResultFromIntent(eventData.intent);` can be null and crash on `googleSignInResult.isSuccess();` call.

I still can't reproduce it but it often comes back in Crashlitics on several Android versions (6, 7, 8, 9).

jpierront:
Fixes #1473


jpierront:
@EddyVerbruggen The Labeler problem is on my side, on your side or on labeler side ? I don't understand ><

EddyVerbruggen:
@jpierront Sorry about that labeler thing. It's a Github Action I was experimenting with.

Thanks for the PR!

PR #1479, opened 2019-11-11
Suggestion: Return error objects instead of error message string

tomcatmwi:
Here's a quick and somewhat quick and dirty implementation for Storage only. If you only return the error message string from Firebase, it's difficult to process in the app. I suggest to return the `exception` or `error` object instead:

```
onFailure: exception => { reject(exception); }
```

That will carry the error code and it can be extracted with `.errorCode()` if needed, and existing code which expects a string also won't break. Seems to work on Android but I didn't test it on iOS.

### DIFF
--- a/src/storage/storage.android.ts
+++ b/src/storage/storage.android.ts
@@ -11,7 +11,7 @@ function getReference(nativeReference: com.google.firebase.storage.StorageRefere
     bucket: nativeReference.getBucket(),
     name: nativeReference.getName(),
     fullPath: nativeReference.getPath(),
-    listAll: (): Promise<ListResult> => listAll({remoteFullPath: nativeReference.getPath(), bucket: listOptions.bucket})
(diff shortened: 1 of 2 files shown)


src/storage/storage.android.ts
```diff
@@ -85,7 +85,7 @@ export function uploadFile(arg: UploadFileOptions): Promise<UploadFileResult> {
 
       const onFailureListener = new gmsTasks.OnFailureListener({
         onFailure: exception => {
-          reject("Upload failed. " + exception);
+          reject(exception);
```
EddyVerbruggen: I'm not sure it's wise to return the native Java Exception object here. It used to be a String representation. Not saying that's perfect, but this is not better imo.

tomcatmwi: I'm not sure either, that's why it's a rough fix. But if not the entire object, the Google error code may still work.

PR #1486, opened 2019-11-19
Firestore - Paginate data with query cursors - code improvements

milansar:
used already existing function toValue to convert javascript values to Java values and mostly formatting correction


EddyVerbruggen:
👍

PR #1490, opened 2019-11-21
removing webinar promo

rdlauer:


EddyVerbruggen:
Thanks Rob - I think I'll add a link to the webinar YT recording instead😄

PR #1508, opened 2019-12-06
fix(android): use application instance from app module

triniwiz:
fixes use cases where a custom application is used since `com.tns.NativeScriptApplication.getInstance()` would be null when it is not the starting class


EddyVerbruggen:
Nice, thanks! I've published a new version.

PR #1516, opened 2019-12-11
Apple Sign-In sometimes fails on iOS

tkrevh:
Apple Sign-In delegate gets garbage collected and login promise is then never resolved.


EddyVerbruggen:
That must have been a pita to figure out. Thank you!!

PR #1517, opened 2019-12-11
Android Apple Sign-In promise is never resolved

tkrevh:
Apple Sign-In promise gets never resolved, because the success listener is missing a resolve call.


EddyVerbruggen:
You're absolutely right, thanks!

PR #1570, opened 2020-03-19
fix link in fb auth docs

rdlauer:
The link provided to configure iOS info.plist wasn't useful, so I updated the instructions a bit.


EddyVerbruggen:
❤️

PR #1572, opened 2020-03-20
Allow mlkit textrecognition to work in landscape mode on iOS

nicolocaruso85:


acrollet:
There is one more fix coming here, to make text recognition work when the device is upside-down.

Issue #1353, opened 2019-08-04
crash while using markingMode: none


patrick-nurt:
Hi Eddy,

After I've changed to using markingMode: none, my app keeps crashing with this error:

```JS: ERROR Error: Uncaught (in promise): EmptyError: no elements in sequence
System.err: com.tns.NativeScriptException: Attempt to use cleared object reference id=5392
System.err: 	at com.tns.Runtime.getJavaObjectByID(Runtime.java:1006)
System.err: 	at com.tns.Runtime.callJSMethodNative(Native Method)
System.err: 	at com.tns.Runtime.dispatchCallJSMethodNative(Runtime.java:1203)
System.err: 	at com.tns.Runtime.callJSMethodImpl(Runtime.java:1083)
System.err: 	at com.tns.Runtime.callJSMethod(Runtime.java:1070)
System.err: 	at com.tns.Runtime.callJSMethod(Runtime.java:1050)
System.err: 	at com.tns.Runtime.callJSMethod(Runtime.java:1042)
System.err: 	at com.tns.gen.com.google.android.gms.tasks.OnCompleteListener.onComplete(OnCompleteListener.java:19)
System.err: 	at com.google.android.gms.tasks.zzj.run(Unknown Source:4)
System.err: 	at android.os.Handler.handleCallback(Handler.java:873)
System.err: 	at android.os.Handler.dispatchMessage(Handler.java:99)
System.err: 	at android.os.Looper.loop(Looper.java:214)
System.err: 	at android.app.ActivityThread.main(ActivityThread.java:7094)
System.err: 	at java.lang.reflect.Method.invoke(Native Method)
System.err: 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
System.err: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)```

patrick-nurt:
I'm using the latest (9.0.4) of the nativescript-plugin-firebase plugin. It looks like this is where it goes bad: 
https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/4c60759c2148fbcbde09f65868de0a828dc8b37c/src/firebase.android.ts#L591

I hope that it can be resolved quickly, as the markingMode really makes a huge difference here.

Best regards

EddyVerbruggen:
[...]

PR #1578, opened 2020-03-31
fix(remote-config): attempt to use cleared object reference

jnorkus:


jnorkus:
#1353

EddyVerbruggen:
Thanks! I smoketested it and it doesn't seem to break the demo app so I'm happy to release this and hope it helps someone.

PR #1579, opened 2020-04-06
Allow different environments with --release tag

allangoncalves:
Currently is not possible to create builds with the auto environment assigning because the `--release` tag always sets the firebase production environment.

To check the current environment, this PR is checking if the env is production and NOT staging/development. Breaking changes are avoided and the users got to use --release with either environment.

Since all versions > 10 of this plugin are supposed to work with NS6.1 or superior, I believe using `--release` to check the environment is redundant since `--release` sets production to true by default.


EddyVerbruggen:
This looks great, thank you!

PR #1639, opened 2020-07-14
ios: remove legacy build system as its no longer needed:

### LINKED ISSUE #1638: iOS Legacy Build System - still necessary?
I've been attempting to test the ios-v8 work that appears to be nearing completion, and have run into an issue during build:
```
error: the use of xcframeworks is not supported in the legacy build system.
```

It looks like `nativescript-plugin-firebase` forces the project to build with this here: https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/publish/scripts/installer.js#L601

The commit associated seems to hint that the firebase team was in a bit of a transition about two years ago: https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/1022

Any idea if using the legacy system is still needed? Removing ios-v8 from the situation, just removing that portion of the after-prepare hook seems to work. It is extremely possible this is unrelated to this project, but with ios-v8 altering the hook similarly made my build work, but then my app would not open (nothing in the build output). You can test the ios v8 work by replacing your tns-ios version with `beta-v8` and doing a clean build from there. It does seem like this question can be answered entirely independently from the v8 work though.

james-criscuolo:
- will be needed for ios-v8
- closes #1638

### DIFF
--- a/publish/scripts/installer.js
+++ b/publish/scripts/installer.js
@@ -253,7 +253,6 @@ function promptQuestionsResult(result) {
     writePodFile(result);
     writeGoogleServiceCopyHook();
     writeBuildscriptHookForCrashlytics(isSelected(result.crashlytics));
-    writeBuildscriptHookForFirestore(isSelected(result.firestore));
     activateIOSCrashlyticsFramework(isSelected(result.crashlytics));
     activateIOSMLKitCameraFramework(isSelected(result.ml_kit));
   }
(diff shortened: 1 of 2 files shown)


EddyVerbruggen:
The only small drawback I see is when folks start using ios-v8 and previously had this plugin with Firestore installed. Then. they have the hook in their hooks folder which still sets the build system to legacy. But I'll take it because the alternative is too much red tape for my taste.

PR #1641, opened 2020-07-19
Disable Direct Channel API on iOS

### LINKED ISSUE #1609: Firebase Messaging iOS Direct Channel Deprecation
Hiya,

I, and I'm sure many others received an email from Firebase today regarding the deprecation of the iOS Direct Channel API. 

> Firebase Messaging iOS Direct Channel has been used for sending data to apps when they are in the foreground and for upstream messages. The Firebase Messaging iOS SDK v7.0 (slated to be released in September 2020) will no longer support the iOS Direct Channel API. Your app will break if you update to the Firebase Messaging iOS SDK v7.0 without migrating off the iOS Direct Channel API.

My question is simply whether this deprecation will affect this plugin and any apps we have built that use it?

Attached: Full email from Firebase
firebase-message

Many Thanks

shahen94:
Direct Channel API is deprecated and will be removed in the next releases of the FirebaseSDK

Same issue on React-Native
https://github.com/invertase/react-native-firebase/issues/3674
https://github.com/invertase/react-native-firebase/pull/3733

Closes #1609

### DIFF
--- a/src/messaging/messaging.ios.ts
+++ b/src/messaging/messaging.ios.ts
@@ -251,10 +251,6 @@ export function prepAppDelegate() {
 
   _addObserver(UIApplicationDidBecomeActiveNotification, appNotification => {
     _processPendingNotifications();
-
-    if (typeof (FIRMessaging) !== "undefined") {
-      FIRMessaging.messaging().shouldEstablishDirectChannel = false;
-    }
   });
 }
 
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
@shahen94 Thanks!

According to https://firebase.google.com/docs/reference/ios/firebasemessaging/api/reference/Classes/FIRMessaging#shouldestablishdirectchannel the `false` / 'NO' value is the default, so you might as well remove those 3 lines entirely.

shahen94:
Good catch! @EddyVerbruggen done

EddyVerbruggen:
Excellent, thank you!

felixkrautschuk:
@EddyVerbruggen is there a plan when this fix will be published? and will that fix be available for the NS 6 compatible version of this plugin?

PR #1684, opened 2020-09-29
Enable BITCODE

mapo80:
This changes permit to enable bitcode on NS apps that use this plugin


abhayastudios:
@mapo80 is this the reason that Xcode doesnt provide me anymore with dSYM?

Issue #1660, opened 2020-09-06
appleSignIn threw " Error: Argument must be an object."


ngocnt1:
NS: 7.0.6 
NS firebase: 11.0.0.rc5
iOS platform:

I got an exception with apple signin, when i tried to deeply debug on plugin code, i realized below is the lines of code has been throwing out the exception.:
**firebase.ios.js** line 824 - 827
` const delegate = ASAuthorizationControllerDelegateImpl.createWithOwnerAndResolveReject(**new WeakRef(this)**, resolve, reject);
                CFRetain(delegate);
                authorizationController.delegate = delegate;
                authorizationController.presentationContextProvider = ASAuthorizationControllerPresentationContextProvidingImpl.createWithOwnerAndCallback(**new WeakRef(this)**);`
(**this** is undefined all the times)

**Workaround**:
`const delegate = ASAuthorizationControllerDelegateImpl.createWithOwnerAndResolveReject(this, resolve, reject);
                CFRetain(delegate);
                authorizationController.delegate = delegate;
                authorizationController.presentationContextProvider = ASAuthorizationControllerPresentationContextProvidingImpl.createWithOwnerAndCallback(this);`

@EddyVerbruggen do we need to pass **this** if now it's **undefined** ?

butaminas:
Any updates on this?

I get the same error when simply trying to Sing In with apple on iOS with Nativescript 7.0.8 (Vue) and @nativescript/firebase 11

ngocnt1:
it still occurs on NS firebase: 11.0.0 release

PR #1686, opened 2020-09-30
Apple sign in for iOS fix / workaround

butaminas:
I don't know if this is the right fix but it doesn't break anything and it makes Apple Sign in work again on iOS.

Since nobody have time to actually investigate and fix this problem I'm making this pull request with suggested fix / workaround from [**ngocnt1**](https://github.com/ngocnt1) in issue #1660 to attract more attention from someone that could actually investigate this problem or at least merge this pull request so people could use Apple Sign in on iOS again.


EddyVerbruggen:
Thanks - I'll take your word for it.

Not sure when a new version is released, but if it is, this will be included.

Issue #1642, opened 2020-07-21
Fabric Crashlytics SDK deprecated


felixkrautschuk:
We got an email some weeks ago informing us that the Fabric Crashlytics SDK is deprecated. Full email:

<img width="577" alt="Bildschirmfoto 2020-07-21 um 11 47 11" src="https://user-images.githubusercontent.com/6443021/88039930-818ff300-cb48-11ea-865c-af262f4d1e66.png">

------------------------

Android upgrade instructions: https://firebase.google.com/docs/crashlytics/upgrade-sdk?platform=android

iOS upgrade instructions: https://firebase.google.com/docs/crashlytics/upgrade-sdk#objective-c_2

mreall:
I'm seeing this message on the Firebase site, too.  Are there plans to upgrade the plugin?

hrueger:
I'm getting the following error:
```
The Android Gradle plugin supports only Crashlytics Gradle plugin version 1.28.0 and higher.
The following dependencies do not satisfy the required version:
root project 'SchoolSquirrel' -> io.fabric.tools:gradle:1.26.1
```
Could this be related somehow?

lambourn:
out of interest: where is the code base for `@nativescript/firebase@rc`? Does not seem to be in this repo, master is still on 10.5.2

jessorlisa:
_Concerning the original topic:_ 
Will the plugin be updated to use the new Firebase SDK? 

Thanks in advance for a quick update.

vstanyshevskyy:
Hi, I got the Fabric deprecation reminder today as well. Is there a plan to address this issue?
Thanks

kanclalg:
This affects us too. Maybe @EddyVerbruggen could share some insights/plans on this one?

EddyVerbruggen:
Of course I also received this notification and numerous questions via mail, twitter, slack, etc. But I’m not done with higher priority stuff yet. Working like crazy to get to this one.

felixkrautschuk:
@EddyVerbruggen thanks for your feedback. I am really appreciating what you are doing (and of course what you already did) for NativeScript. Your plugins play a very important role for this framework.

000panther:
[...]

PR #1699, opened 2020-10-11
Upgrade Fabric/Crashlytics to new FirebaseCrashlytics

kl4n4:
As discussed in #1642 Google is deprecating the Fabric Crashlytics SDK on November 15, 2020 so to continue the crashlytics support for this plugin it has to be [upgraded to the new FirebaseCrashlytics SDK](https://firebase.google.com/docs/crashlytics/upgrade-sdk)

My colleague @000panther and myself upgraded the Android and iOS specific SDKs to the new dependencies and new API. For iOS I also was able to remove the custom wrapper library `TNSCrashlyticsLogger` since it is no longer needed - just as its creator @kanclalg suggested. 

### Known issues/open tasks:

- [ ] *iOS*: Manually forcing a crash is not implemented (https://github.com/EddyVerbruggen/nativescript-plugin-firebase/commit/05186eedd052f64021ce90395bb18f50eeed9cc1#diff-397fe7e75702bf1ec8281547b4014e75R57). Since I am not a native ObjectiveC developer and I could not figure out a way how to implement [this](https://firebase.google.com/docs/crashlytics/test-implementation?authuser=0&platform=ios#force_a_crash_to_test_your_implementation) with NativeScript JS

Please feel free to test it yourselves and give us feedback.


src/firebase.ios.ts
```diff
@@ -344,9 +344,9 @@ firebase.init = arg => {
         }
       }
 
-      if (arg.crashlyticsCollectionEnabled && typeof (Crashlytics) !== "undefined") {
-        Fabric.with(NSArray.arrayWithObject(Crashlytics.class()));
-      }
+      // if (arg.crashlyticsCollectionEnabled && typeof (Crashlytics) !== "undefined") {
```
Codex-, line 347: Should this be removed?

kl4n4: yes, you are right I unfortunately missed to clean up that code. 
For [reference](https://firebase.google.com/docs/crashlytics/upgrade-sdk?authuser=0&platform=ios#firebasecrashlytics_no_longer_works_with_the_fabric_sdk): now only `FIRApp.configure()` should be necessary to initialize the new crashlytics SDK (which is done a few lines above).

src/crashlytics/crashlytics.android.ts
```diff
@@ -5,72 +5,69 @@ declare const com: any;
...
+      const fullMessage = priority + '/' + tag + ': ' + msg;
+      com.google.firebase.crashlytics.FirebaseCrashlytics.getInstance().log(fullMessage);
     } else {
-      com.crashlytics.android.Crashlytics.log(msg);
+      com.google.firebase.crashlytics.FirebaseCrashlytics.getInstance().log(msg);
     }
   }
 }
 
 export function setString(key: string, value: string): void {
```
Codex-, line 23: Since all these methods deeply call the same method, perhaps mark the parent typings for these methods as deprecated and add a singular function to replace them? Would make it easier to digest / consume

000panther: The reason that I did not do what you recommended was to keep the change size and the impact on the existing api minimal, to have a quick solution that can be used on 15th of November. I think that this refactoring should be done, but maybe not right now.

Codex-: This would only add to the API, not remove used API's, marking deprecated just prepares people for the future removal is all.

src/crashlytics/crashlytics.android.ts
```diff
@@ -5,72 +5,69 @@ declare const com: any;
```
Codex-, line 5: `import * as appModule from 'tns-core-modules/application';` should be removed, not ns7 compat

src/crashlytics/crashlytics.ios.ts
```diff
@@ -2,71 +2,71 @@ import { ENABLE_CRASHLYTICS_HINT } from "./crashlytics-common";
...
   if (isCrashlyticsAvailable()) {
-    Crashlytics.sharedInstance().setUserIdentifier(id);
+    FIRCrashlytics.crashlytics().setUserID(id);
   }
 }
 
 export function crash(): void {
   if (isCrashlyticsAvailable()) {
-    Crashlytics.sharedInstance().crash();
+    // TODO: manually force crash - https://firebase.google.com/docs/crashlytics/test-implementation?authuser=0&platform=ios#force_a_crash_to_test_your_implementation
```
Codex-, line 57: From my understanding, you can crash with:
```ts
    NSException.exceptionWithNameReasonUserInfo(
      "FIRCrashlytics",
      "test crash",
      null
    ).raise();
```

I tested this in a demo app I made quickly (since the one in this repo isn't working for me):
```
***** Fatal JavaScript exception - application has been terminated. *****
NativeScript encountered a fatal error: Uncaught Error: test crash

...

*** First throw call stack:
...
```

EddyVerbruggen:
Many thanks to all involved! ❤️

I've just published 10.6.0 to npm which includes these changes.

I'm now going to attempt to make this {N}7 compatible as well.

PR #1718, opened 2020-11-03
Fix bug on iOS platform occurring when application is built in releas…

Execomsrl:
Fix bug on iOS platform occurring when application is built in release mode with development environment. In this case, before this patch, the plugin copies firebase prod file in place of the firebase dev file.

### DIFF
--- a/publish/scripts/installer.js
+++ b/publish/scripts/installer.js
@@ -846,7 +846,7 @@ module.exports = function($logger, hookArgs) {
             });
         }
 
-        var buildType = isReleaseBuild || isProdEnv ? 'production' : 'development';
+        var buildType = (isReleaseBuild && isProdEnv) || (!isReleaseBuild && isProdEnv) ? 'production' : 'development';
 
         /*
             Detect if we have nativescript-plugin-firebase temp file created during after-prepare hook, so we know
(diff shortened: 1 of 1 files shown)


publish/scripts/installer.js
```diff
@@ -846,7 +846,7 @@ module.exports = function($logger, hookArgs) {
             });
         }
 
-        var buildType = isReleaseBuild || isProdEnv ? 'production' : 'development';
+        var buildType = (isReleaseBuild && isProdEnv) || (!isReleaseBuild && isProdEnv) ? 'production' : 'development';
```
EddyVerbruggen, line 850: Seems logically equivalent to `var buildType = isProdEnv ? 'production' : 'development';`

Execomsrl: Yes you are right i'm adding some code to make it work as desired and send a new pull request

mapo80:
We need this fix to be merged! Good fix

mapo80:
Hi @EddyVerbruggen, thanks for accepting this PR. Do you plan to release it on 10.6.x version. I hope so.

Thanks very much.

Execomsrl:
Thanks for merging the fix. Can you apply it to 10.6 version too? I'm actually using that version and it would be great if i could use the fix there too. Thanks.

EddyVerbruggen:
For 10.6 you should point a PR to the 10.6 maintenance branch ;)
I'd be happy to merge it of course.

Execomsrl:
> For 10.6 you should point a PR to the 10.6 maintenance branch ;)
> I'd be happy to merge it of course.

I sent a new PR to fix the bug even in 10.6 branch. Thanks.

PR #1852, opened 2021-10-15
Update Facebook SDK to latest

abhayastudios:
Android version was pinned to v5 whereas Facebook recently requires a more recent version (latest major is v12) due to security considerations where the previously used webview is not allowed anymore and prevents Android users from using FB auth.


brendonwm:
https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/1858

delaneyb:
facebook-login v12 introduces a transitive dependency via facebook-common on androidx.fragment 1.3.0 that breaks navigation in nested frames, as @nativescript/core and @nativescript/android expect 1.2.5 which doesn't cause the issue. Here is an alternate PR for 11.+: https://github.com/EddyVerbruggen/nativescript-plugin-firebase/pull/1868 that still resolves the webview issue but doesn't have the fragment dep.

PR #1868, opened 2022-01-02
Update facebook-login sdk dep, add warning about facebook-common dep

delaneyb:
This PR updates the facebook-login android dependency to 11.+. There is no need to separately include facebook-core as it is a subdependency of facebook-login (see https://developers.facebook.com/docs/android/componentsdks & https://mvnrepository.com/artifact/com.facebook.android/facebook-login/11.3.0).

Importantly, the facebook-login version as been restricted to 11.+, as from 12.0.0, its subdependency facebook-common adds a dependency on androidx.fragment 1.3.0, overriding the @nativescript-android's expected ver 1.2.5, which breaks navigation in nested frames - minimal repro here: https://github.com/delaneyb/androidx-fragment-130-repro.

### DIFF
--- a/publish/scripts/installer.js
+++ b/publish/scripts/installer.js
@@ -682,8 +682,17 @@ dependencies {
     ` + (isSelected(result.ml_kit_natural_language_smartreply) ? `` : `//`) + ` implementation "com.google.firebase:firebase-ml-natural-language-smart-reply-model:20.0.7"
 
     // Facebook Authentication
-    ` + (isSelected(result.facebook_auth) ? `` : `//`) + ` implementation "com.facebook.android:facebook-core:5.4.0"
-    ` + (isSelected(result.facebook_auth) ? `` : `//`) + ` implementation "com.facebook.android:facebook-login:5.4.0"
+    
+    // IMPORTANT: facebook-login SDK release versions are synced with its dependency package
+    // "facebook-common", which in 12.0.0 introduces a dependency on androidx.fragment:fragment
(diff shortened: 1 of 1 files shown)


brendonwm:
Hi @delaneyb . Is there a reasonable way to doctor my local setup to use your _installer.js_ change? We don't really want to upgrade NativeScript versions at the moment (we're on 6.5.3) to use their included Firebase features right now.

(Also looking into using firebase npm package directly - but that's intended for web and I'm not sure it'll work with an NS app.)

Thanks.

delaneyb:
> Hi @delaneyb . Is there a reasonable way to doctor my local setup to use your _installer.js_ change? We don't really want to upgrade NativeScript versions at the moment (we're on 6.5.3) to use their included Firebase features right now.
> 
> (Also looking into using firebase npm package directly - but that's intended for web and I'm not sure it'll work with an NS app.)
> 
> Thanks.

Hi @brendonwm,

You could look at using https://www.npmjs.com/package/patch-package if you don't want to clone/fork and build the plugin. Unfortunately patch-package doesn't play nice with Yarn 2 and I always ran up against issues with Yarn's own `yarn patch` command, so if you are in that boat you are probably best to clone https://github.com/delaneyb/nativescript-plugin-firebase/tree/update-facebook-login, package it to a .tgz (publish/pack.sh?) and reference that from your apps package.json with a file path.

OPADA-Eng:
Any progress here ?

